PR #12, opened 2018-12-25
Added CreateCustomer example.

fiboknacky:


examples/AccountManagement/CreateCustomer.php
```diff
@@ -0,0 +1,134 @@
...
+
+/**
+ * This example illustrates how to create a new customer under a given manager account.
+ *
+ * Note: this example must be run using the credentials of a Google Ads manager account. By default,
+ * the new account will only be accessible via the manager account.
+ */
+class CreateCustomer
+{
+    const CUSTOMER_ID = 'INSERT_CUSTOMER_ID_HERE';
```
jimper: Shouldn't you use a different constant, since this needs to be a manager ID specifically? Also a little confusing to require a customer ID to create a customer.

fiboknacky: Thanks for catching this.
Actually the ID shouldn't be specified in the new Customer object too. I'll point this out in the golden PR.

jimper (approved):
LGTM

sundquist (approved):
LGTM

PR #15, opened 2019-01-09
Added GetHotelAdsPerformance example

fiboknacky:


examples/Reporting/GetHotelAdsPerformance.php
```diff
@@ -0,0 +1,143 @@
...
+     * @param int $customerId the client customer ID without hyphens
+     */
+    public static function runExample(GoogleAdsClient $googleAdsClient, $customerId)
+    {
+        $googleAdsServiceClient = $googleAdsClient->getGoogleAdsServiceClient();
+        // Creates a query that retrieves hotel-ads statistics for each campaign and ad group.
+        // Returned statistics will be segmented by the check-in day of week and length of stay.
+        $query =
+            "SELECT campaign.id, "
+                . "campaign.advertising_channel_type, "
```
jimper: Do you need to select campaign.advertising_channel_type and ad_group.status if you aren't printing them? Seems they're only used in the where clause.

examples/Reporting/GetHotelAdsPerformance.php
```diff
@@ -0,0 +1,143 @@
...
+                . "ad_group.id, "
+                . "ad_group.status, "
+                . "metrics.impressions, "
+                . "metrics.clicks, "
```
jimper: This doesn't seem to be used at all

examples/Reporting/GetHotelAdsPerformance.php
```diff
@@ -0,0 +1,143 @@
...
+        $response =
+            $googleAdsServiceClient->search($customerId, $query, ['pageSize' => self::PAGE_SIZE]);
+
+        // Iterates over all rows in all pages and prints the requested field values for each row.
+        foreach ($response->iterateAllElements() as $googleAdsRow) {
+            /** @var GoogleAdsRow $googleAdsRow */
+            // Note: A mapping of enum names to values for days of week can be found at
+            // DayOfWeek.php.
+            printf(
+                "Ad group with ID %d in campaign with ID %d "
```
jimper: Trying to fit this much info into a single sentence is a little cumbersome. I think that breaking this up into multiple lines would make it easier to parse. If you really want to keep it all on one line, at least consider simplifying:

"Ad group ID %d in campaign ID %d with hotel check-in on %d and %d days of stay..."

examples/Reporting/GetHotelAdsPerformance.php
fiboknacky: It's required to be in here if they will be used in the where clause.

examples/Reporting/GetHotelAdsPerformance.php
fiboknacky: Do you mean breaking the "code" into multiple lines or both the "code" and the printed "message" ?
Anyway, I've revised the message per your suggestion.

examples/Reporting/GetHotelAdsPerformance.php
fiboknacky: Done.

examples/Reporting/GetHotelAdsPerformance.php
jimper: I meant breaking up the printed message itself into multiple lines, to make it easier for users to parse. Personally I think you're trying to cram too much into a single sentence which makes it hard to parse.

Alternatively you could consider restricting the select more and cut out some of the metrics being printed. Eg, filter by a specific check-in day and/or length of stay so it doesn't need to be printed for each result.

But if you really want to include everything I won't stop you. Just please remove the first 2 "with"s since the current with-with-with structure doesn't read well :)

fiboknacky: Done removing the first "with"s.
I've decided to do it this way, as this is quite consistent with what we've done in other examples.
I really need check-in day and length of stay, as they are both important segment fields that we want to show. Requiring 2 attribute fields (campaign ID and ad group ID), 2 segment fields (check-in day and length of stay) and 2 metric fields seems reasonable to me.

Besides, the printed message is broken into many lines based on the above-mentioned categories of fields. I think it should be readable to some extent.

AnashOommen (approved):

jimper (approved):
LGTM

PR #23, opened 2019-02-09
Updated getopt dependency to the latest version

Raibaz:
This also fixes a test that fails with getopt  v3.2.2 but passes with v3.0.

### DIFF
--- a/tests/Google/Ads/GoogleAds/Examples/Utils/ArgumentParserTest.php
+++ b/tests/Google/Ads/GoogleAds/Examples/Utils/ArgumentParserTest.php
@@ -97,7 +97,7 @@ class ArgumentParserTest extends TestCase
     public function testPassingRequiredArgumentButMissingValue()
     {
         $this->expectOutputRegex('/Usage/');
-        $_SERVER['argv'] = [null, '--customerId', '--campaignId', 11111];
+        $_SERVER['argv'] = [null, '--campaignId', 11111, '--customerId'];
         $this->argumentParser->parseCommandArguments(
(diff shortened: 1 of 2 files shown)


tests/Google/Ads/GoogleAds/Examples/Utils/ArgumentParserTest.php
```diff
@@ -97,7 +97,7 @@ public function testPassingRequiredAndOptionalArgumentsWhenOptionalArgumentMissi
     public function testPassingRequiredArgumentButMissingValue()
     {
         $this->expectOutputRegex('/Usage/');
-        $_SERVER['argv'] = [null, '--customerId', '--campaignId', 11111];
+        $_SERVER['argv'] = [null, '--campaignId', 11111, '--customerId'];
```
fiboknacky: This shouldn't be changed.
The test is aimed to test the case when customerId (which is required) is specified, but no value is given.

Raibaz: You're right, but getopt changed its behavior between versions and now interprets '--campaignId' as the value for '--customerId', so to test the case when customerId is specified but no value is given this is correct.

Do you think we should add a specific test for the case when an example is invoked with something like `--customerId --campaignId 111111` or should we just leave it as it is, having examples fail with `request_error: Invalid customer ID '--campaignId'`?

fiboknacky: Thanks for confirming. Then, this already LGTM.

PR #30, opened 2019-03-01
Fixed access to the hotel_length_of_stay segment in the response

Raibaz:


examples/Reporting/GetHotelAdsPerformance.php
```diff
@@ -130,7 +130,7 @@ public static function runExample(GoogleAdsClient $googleAdsClient, $customerId)
                 $googleAdsRow->getAdGroup()->getId()->getValue(),
                 $googleAdsRow->getCampaign()->getId()->getValue(),
                 $googleAdsRow->getHotelCheckInDayOfWeek(),
```
fiboknacky: This too. I think your IDE  should warn this, no?

Raibaz: Done.

PR #33, opened 2019-03-07
Do not require ReactPHP HTTP library 

alexeyshockov:
`"react/http"` is required only `AuthenticateInWebApplication.php` example and is not used by any real library code.

Requiring it doesn't make sense for production usage, so looks like it should be optional.

### DIFF
--- a/examples/Authentication/AuthenticateInWebApplication.php
+++ b/examples/Authentication/AuthenticateInWebApplication.php
@@ -59,6 +59,11 @@ class AuthenticateInWebApplication
 
     public static function main()
     {
+        if (!class_exists(Server::class)) {
+            echo 'Please install "react/http" package to be able to run this example';
+            exit(1);
+        }
+
         $loop = Factory::create();
         // Creates a socket for localhost with random port.
         $socket = new \React\Socket\Server(0, $loop);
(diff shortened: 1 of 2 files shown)


fiboknacky:
Could you please sign CLA too?
Thanks.

alexeyshockov:
@fiboknacky, done

Also added a class check in the example

examples/Authentication/AuthenticateInWebApplication.php
```diff
@@ -59,6 +59,11 @@ class AuthenticateInWebApplication
 
     public static function main()
     {
+        if (!class_exists(Server::class)) {
```
fiboknacky: Can you refer to the class using FQCN?

alexeyshockov: IMO you import this class anyway (to use it later), so no need for FQCN. It also works without any problems when the library is not loaded and the class is not available.

But if you really want the change, I will do it, of course.

fiboknacky: That's fine. Yes, I already did the import above in this class. :)

alexeyshockov:
@fiboknacky, @Raibaz, any insights on when it could be merged? I signed the CLA, so all seems OK now.

And thanks for the great workshop in Munich, BTW :)

fiboknacky:
Sorry for the late reply. Will do now.
Thanks for joining us in the event too. :)

PR #36, opened 2019-03-14
Example and utility for partial failure

Raibaz:


examples/BasicOperations/HandlePartialFailure.php
```diff
@@ -0,0 +1,201 @@
...
+use GetOpt\GetOpt;
+use Google\Ads\GoogleAds\Examples\Utils\ArgumentNames;
+use Google\Ads\GoogleAds\Examples\Utils\ArgumentParser;
+use Google\Ads\GoogleAds\Lib\V1\GoogleAdsClient;
+use Google\Ads\GoogleAds\Lib\V1\GoogleAdsClientBuilder;
+use Google\Ads\GoogleAds\Lib\V1\GoogleAdsException;
+use Google\Ads\GoogleAds\Lib\OAuth2TokenBuilder;
+use Google\Ads\GoogleAds\Util\V1\ResourceNames;
+use Google\Ads\GoogleAds\V1\Errors\GoogleAdsError;
+//use Google\Ads\GoogleAds\V1\Services\GoogleAdsRow;
```
fiboknacky: Remove

examples/BasicOperations/HandlePartialFailure.php
```diff
@@ -0,0 +1,201 @@
...
+     * Runs the example.
+     *
+     * @param GoogleAdsClient $googleAdsClient the Google Ads API client
+     * @param int $customerId the client customer ID without hyphens
+     * @param int $campaignId a campaign ID
+     */
+    public static function runExample(GoogleAdsClient $googleAdsClient, $customerId, $campaignId)
+    {
+    
+        $temp_unused_var = new \Google\Ads\GoogleAds\V1\Errors\GoogleAdsFailure();
```
fiboknacky: Use camelCase + do import of GoogleAdsFailure

examples/BasicOperations/HandlePartialFailure.php
```diff
@@ -0,0 +1,201 @@
...
+
+        $campaignResourceName =
+            new StringValue(['value' => ResourceNames::forCampaign($customerId, $campaignId)]);
+
+        // This AdGroup should be created successfully - assuming the campaign in the params
```
fiboknacky: Nit: This "ad group"

examples/BasicOperations/HandlePartialFailure.php
```diff
@@ -0,0 +1,201 @@
...
+            new StringValue(['value' => ResourceNames::forCampaign($customerId, $campaignId)]);
+
+        // This AdGroup should be created successfully - assuming the campaign in the params 
+        // exists.
+        $adGroup1 = new AdGroup([
+            'name' => new StringValue(['value' => 'Valid AdGroup #' . uniqid()]),
+            'campaign' => $campaignResourceName            
+        ]);
+
+        // This AdGroup will always fail - campaign ID 0 in resource names is never valid.
```
fiboknacky: campaign ID 0 in "the resource name" is never valid.

examples/BasicOperations/HandlePartialFailure.php
fiboknacky: Please fix all below occurrences too.

examples/BasicOperations/HandlePartialFailure.php
```diff
@@ -0,0 +1,201 @@
...
+        {
+            printf(
+                "All operations completed successfully. No partial failures to show.%s",
+                PHP_EOL
+            );
+        }
+
+        // Finds the failed operations by looping through the results.
+        $operationIndex = 0;
+        foreach($response->getResults() as $result)
```
fiboknacky: Fix all PSR-2 problems here.
Always run phpcs first please.

examples/BasicOperations/HandlePartialFailure.php
```diff
@@ -0,0 +1,201 @@
...
+
+        // Issues the mutate request, setting partialFailure=true.
+        $adGroupServiceClient = $googleAdsClient->getAdGroupServiceClient();
+        $response = $adGroupServiceClient->mutateAdGroups(
+            $customerId,
+            $operations,
+            ['partialFailure' => true]
+        );
+
+        if(!is_null($response->getPartialFailureError()))
```
fiboknacky: This clause is not PSR-2 compliant.

src/Google/Ads/GoogleAds/V0/Util/ErrorUtils.php
```diff
@@ -0,0 +1,164 @@
...
+
+    /**
+     * Checks if a result in a mutate response is a partial failure.
+     *
+     * @param \Google\Protobuf\Internal\Message $message
+     * @return bool
+     */
+    public static function isPartialFailure(Message $message)
+    {
+        return $message->byteSize() == 0;
```
fiboknacky: Use ===

src/Google/Ads/GoogleAds/V0/Util/ErrorUtils.php
```diff
@@ -0,0 +1,164 @@
...
+    }
+ }
```
fiboknacky: Add another blank line pls.

src/Google/Ads/GoogleAds/V0/Util/ErrorUtils.php
```diff
@@ -0,0 +1,164 @@
+<?php
```
fiboknacky: Path is wrong. Please check with the existing utils.

examples/BasicOperations/HandlePartialFailure.php
fiboknacky: And this shouldn't be here.
It should be at least very close to when we call mutateAdGroups and you need to explain why we need this.

It'd be better to make your ErrorUtils instantiable and include this line in the constructor instead.
Users wouldn't need to know about this detail.
You can provide this as a constructor parameter to help dependency injection.

src/Google/Ads/GoogleAds/V0/Util/ErrorUtils.php
fiboknacky: Please add test too.

src/Google/Ads/GoogleAds/V0/Util/ErrorUtils.php
```diff
@@ -0,0 +1,164 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Ads\GoogleAds\V0\Util;
+
+use \Google\Protobuf\Any;
```
fiboknacky: Prefix \ is not necessary.

src/Google/Ads/GoogleAds/V0/Util/ErrorUtils.php
```diff
@@ -0,0 +1,164 @@
...
+     *         }
+     *       }
+     *     }
+     *   </code>
+     * </pre>
+     *
+     * A single GoogleAdsError instance would be returned for operation index 1 and 2, and an empty
+     * list otherwise.
+     *
+     * @param int operationIndex the index of the operation, starting from 0.
```
fiboknacky: Nit: we don't end with a period for @param and @return

src/Google/Ads/GoogleAds/V0/Util/ErrorUtils.php
```diff
@@ -0,0 +1,164 @@
...
+     * @param \Google\Rpc\Status partialFailureStatus a partialFailure status, with the detail list
```
fiboknacky: Include Status and don't use the full path here.

src/Google/Ads/GoogleAds/V0/Util/ErrorUtils.php
```diff
@@ -0,0 +1,164 @@
...
+     *     }
+     *   </code>
+     * </pre>
+     *
+     * A single GoogleAdsError instance would be returned for operation index 1 and 2, and an empty
+     * list otherwise.
+     *
+     * @param int operationIndex the index of the operation, starting from 0.
+     * @param \Google\Rpc\Status partialFailureStatus a partialFailure status, with the detail list 
+     *  containing GoogleAdsFailure instances.
```
fiboknacky: Need to indent for 4 spaces for hanging indentation.

src/Google/Ads/GoogleAds/V0/Util/ErrorUtils.php
```diff
@@ -0,0 +1,164 @@
...
+        $result = [];
+        foreach($failure->getErrors() as $error)
+        {
+            $pathElements = $error->getLocation()->getFieldPathElements();
+            if(count($pathElements) > 0)
+            {
+                $element = $pathElements[0];
+                $fieldName = $element->getFieldName();
+                $index = $element->getIndex();
+                if($fieldName == "operations" && $index->getValue() == $operationIndex)
```
fiboknacky: Use ===

src/Google/Ads/GoogleAds/V0/Util/ErrorUtils.php
```diff
@@ -0,0 +1,164 @@
...
+            $errors = self::getGoogleAdsErrorsFromFailure($operationIndex, $failure);
+            $result = array_merge($result, $errors);
+        }
+        return $result;
+    }
+
+    /**
+     * Return a list of GoogleAdsError instances for a given operation index.
+     *
+     * @see #getGoogleAdsErrorsFromStatus
```
fiboknacky: @see ErrorUtils::getGoogleAdsErrorsFromStatus

src/Google/Ads/GoogleAds/V0/Util/ErrorUtils.php
```diff
@@ -0,0 +1,164 @@
...
+     * list otherwise.
+     *
+     * @param int operationIndex the index of the operation, starting from 0.
+     * @param \Google\Rpc\Status partialFailureStatus a partialFailure status, with the detail list 
+     *  containing GoogleAdsFailure instances.
+     * @return Google\Ads\GoogleAds\V1\Errors\GoogleAdsError[] an array containing the 
+     *  GoogleAdsError instances for a given operation index.
+     */
+    public static function getGoogleAdsErrorsFromStatus(
+        int $operationIndex,
```
fiboknacky: Is this a valid PHP 7.1 syntax?

fiboknacky (commented):
- Check your code styles in all places. It should comply with PSR-2. Always run phpcs before exporting PRs.
- Please apply the same comment to v1 util file too.
- Name of the utils would need a change. We stick to the Guava naming style, e.g., ResourceNames.

examples/BasicOperations/HandlePartialFailure.php
Raibaz: I don't like much the idea of having the ErrorUtils class instantiable, but I agree that that initialization should not be there. I made an `ErrorUtils::initialize()` function that does this as close as possible to the mutate call but still allows to keep all functions static, as the ErrorUtils class does not in fact have any state.

WDYT?

src/Google/Ads/GoogleAds/V0/Util/ErrorUtils.php
Raibaz: Yep scalar type hinting has been introduced in PHP 7: http://php.net/manual/en/migration70.new-features.php

examples/BasicOperations/HandlePartialFailure.php
fiboknacky: Could you elaborate why you don't like it? Actually, it's used in that way in many of our PHP SOAP library utilities.
Being non-static actually helps when you write test too, as you don't need to worry about inadvertently touch the same parts inadvertently.

BTW, having object instantiation in the initialize() method smells like that we actually need a constructor too.

Raibaz: I don't like it because, although it contains an object instantiation, the ErrorUtils does not really have a state itself, so all the methods already in there can be used statically (and, in fact, they are).

The reason why we need to instantiate a GoogleAdsFailure is because of a quirk with the descriptor pool, not because the ErrorUtils class actually uses it or it keeps track of a GoogleAdsFailure object as its state.

As an alternative approach, what if we did the GoogleAdsFailure instantiation inside the `getGoogleAdsFailureFromAny` function in ErrorUtils?

It's the closest place to where it's actually needed and it would be transparent for clients of the ErrorUtils class, without needing to instantiate it or to call `initialize()`.

fiboknacky: > I don't like it because, although it contains an object instantiation, the ErrorUtils does not really have a state itself, so all the methods already in there can be used statically (and, in fact, they are).

> The reason why we need to instantiate a GoogleAdsFailure is because of a quirk with the descriptor pool, not because the ErrorUtils class actually uses it or it keeps track of a GoogleAdsFailure object as its state.

That's a fair point. I'm fine with that too.

Note that instantiation of GoogleAdsFailure needs to be done only before calling the method that uses partial failure.
So, getGoogleAdsFailureFromAny() cannot be used in this case.
initialize() seems the closest we can get.

Regarding the class structure, can you follow the style I said before?
Other examples than ResourceNames would be [CsvFiles](https://github.com/googleads/googleads-php-lib/blob/master/src/Google/AdsApi/AdManager/Util/v201902/CsvFiles.php) and [AdManagerDates](https://github.com/googleads/googleads-php-lib/blob/master/src/Google/AdsApi/AdManager/Util/v201902/AdManagerDates.php).

After looking at your methods, I think we should divide them into groups:

- PartialFailures contains initialize() (to instantiate GoogleAdsFailure)
- GoogleAdsError contains fromStatus() and fromFailure()
- GoogleAdsFailure contains fromAny() and fromStatus()

Make sure you add tests for them too.

Raibaz: Right, I split the ErrorUtils class like you mentioned ad added test coverage.

examples/BasicOperations/HandlePartialFailure.php
```diff
@@ -0,0 +1,194 @@
...
+        $response = $adGroupServiceClient->mutateAdGroups(
+            $customerId,
+            $operations,
+            ['partialFailure' => true]
+        );
+
+        if (!is_null($response->getPartialFailureError())) {
+            printf("Partial failures occurred. Details will be shown below.%s", PHP_EOL);
+        } else {
+            printf(
```
fiboknacky: This command has over 100 chars?

examples/BasicOperations/HandlePartialFailure.php
```diff
@@ -0,0 +1,194 @@
...
+                "All operations completed successfully. No partial failures to show.%s",
+                PHP_EOL
+            );
+        }
+
+        // Finds the failed operations by looping through the results.
+        $operationIndex = 0;
+        foreach ($response->getResults() as $result) {
+            if (PartialFailures::isPartialFailure($result)) {
+                $errors = GoogleAdsErrors::fromStatus(
```
fiboknacky: Can this fit in one line?

src/Google/Ads/GoogleAds/V0/Util/GoogleAdsErrors.php
```diff
@@ -0,0 +1,118 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Ads\GoogleAds\V0\Util;
+
+use \Google\Rpc\Status;
```
fiboknacky: Preceding \ wouldn't be needed.

src/Google/Ads/GoogleAds/V0/Util/GoogleAdsErrors.php
```diff
@@ -0,0 +1,118 @@
+<?php
```
fiboknacky: You still put the file in the wrong directory.
It must be GoogleAds/Util/V0.

src/Google/Ads/GoogleAds/V0/Util/GoogleAdsErrors.php
```diff
@@ -0,0 +1,118 @@
...
+     *     }
+     *   </code>
+     * </pre>
+     *
+     * A single GoogleAdsError instance would be returned for operation index 1 and 2, and an empty
+     * list otherwise.
+     *
+     * @param int operationIndex the index of the operation, starting from 0.
+     * @param Status partialFailureStatus a partialFailure status, with the detail list
+     *    containing GoogleAdsFailure instances.
```
fiboknacky: Again, @param and @return don't end with a period.

fiboknacky: And you need to start at hanging indented line at char 5, not char 4

src/Google/Ads/GoogleAds/V0/Util/GoogleAdsErrors.php
```diff
@@ -0,0 +1,118 @@
...
+     * @return GoogleAdsError[] an array containing the
```
fiboknacky: Do you really need to indent this line?
Or at least, we should indent at later positions. This is a bit too early.

src/Google/Ads/GoogleAds/V0/Util/GoogleAdsErrors.php
```diff
@@ -0,0 +1,118 @@
...
+        foreach ($partialFailureStatus->getDetails() as $detail) {
+            $failure = GoogleAdsFailures::fromAny($detail);
+            $errors = self::fromFailure($operationIndex, $failure);
+            $result = array_merge($result, $errors);
+        }
+        return $result;
+    }
+
+    /**
+     * Return a list of GoogleAdsError instances for a given operation index.
```
fiboknacky: "Returns"

src/Google/Ads/GoogleAds/V0/Util/GoogleAdsErrors.php
```diff
@@ -0,0 +1,118 @@
...
+    ) {
+        $result = [];
+        foreach ($failure->getErrors() as $error) {
+            $pathElements = $error->getLocation()->getFieldPathElements();
+            if (count($pathElements) > 0) {
+                /** @var FieldPathElement $element */
+                $element = $pathElements[0];
+                $fieldName = $element->getFieldName();
+                $index = $element->getIndex();
+                if ($fieldName === "operations" && $index->getValue() == $operationIndex) {
```
fiboknacky: Use === everywhere when applicable.

src/Google/Ads/GoogleAds/V0/Util/GoogleAdsFailures.php
```diff
@@ -0,0 +1,51 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Ads\GoogleAds\V0\Util;
+
+use \Google\Protobuf\Any;
```
fiboknacky: No \

src/Google/Ads/GoogleAds/V0/Util/GoogleAdsFailures.php
```diff
@@ -0,0 +1,51 @@
...
+use \Google\Rpc\Status;
+use Google\Ads\GoogleAds\V0\Errors\GoogleAdsFailure;
+
+final class GoogleAdsFailures
+{
+    /**
+     * Unpacks a single GoogleAdsFailure from an Any instance.
+     *
+     * @param Any An Any instance to unpack
```
fiboknacky: Don't need to capitalize the "An"

src/Google/Ads/GoogleAds/V0/Util/PartialFailures.php
```diff
@@ -0,0 +1,43 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Ads\GoogleAds\V0\Util;
+
+use \Google\Protobuf\Internal\Message;
```
fiboknacky: No \

src/Google/Ads/GoogleAds/V0/Util/PartialFailures.php
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Checks if a result in a mutate response is a partial failure.
+     *
+     * @param Message $message
+     * @return bool
+     */
+    public static function isPartialFailure(Message $message)
+    {
+        return $message->byteSize() == 0;
```
fiboknacky: I was told by the protobuf lib owner that it's not a good idea to use byteSize(), as the c extension doesn't have it.
He suggested to use serialize() and then check for its size instead.

tests/Google/Ads/GoogleAds/V1/Util/PartialFailuresTest.php
```diff
@@ -0,0 +1,56 @@
...
+ * Unit tests for `PartialFailures`.
+ *
+ * @see PartialFailures
+ * @small
+ */
+class PartialFailuresTest extends TestCase
+{
+    public function testIsPartialFailure()
+    {
+        $messageMock = $this
```
fiboknacky: Instead of mocking many thing, can you research to see if we can just instantiate a Message?
We should avoid doing "over-mocking", when possible.

fiboknacky (changes requested):

src/Google/Ads/GoogleAds/V0/Util/PartialFailures.php
Raibaz: Done.

src/Google/Ads/GoogleAds/V0/Util/GoogleAdsErrors.php
Raibaz: Sorry, I was mistaken, I just realized that this folder was the one for generated code. Fixed it.

tests/Google/Ads/GoogleAds/V1/Util/PartialFailuresTest.php
Raibaz: Agreed, I replaced mocks with actual object instances in most cases. I left the mocks where creating an object with the expected behavior was not immediate, such as driving `any->unpack` or message serialization.

examples/BasicOperations/HandlePartialFailure.php
Raibaz: Nope :(

.gitignore
```diff
@@ -1,3 +1,4 @@
 .idea
```
fiboknacky: This file shouldn't be part of this PR.

tests/Google/Ads/GoogleAds/V0/Util/GoogleAdsFailuresTest.php
```diff
@@ -15,70 +15,46 @@
...
-        $emptyStatusMock = $this
-            ->getMockBuilder(Status::class)
-            ->disableOriginalConstructor()
-            ->getMock();
-        $emptyStatusMock
-            ->method("getDetails")
-            ->willReturn([]);
-
-        $this->assertCount(0, GoogleAdsFailures::fromStatus($emptyStatusMock));
+        $emptyStatus = new Status();
```
fiboknacky: Nit: this assignment wouldn't be necessary.

src/Google/Ads/GoogleAds/Util/V0/PartialFailures.php
```diff
@@ -0,0 +1,43 @@
...
+use Google\Ads\GoogleAds\V0\Errors\GoogleAdsFailure;
+
+final class PartialFailures
+{
+    /**
+     * Initializes a GoogleAdsFailure to make sure that the metadata pool knows about it.
+     */
+    public static function initialize()
+    {
+        $tempUnusedVar = new GoogleAdsFailure();
```
fiboknacky: Heads-up: This wouldn't be needed once #41 is merged.

fiboknacky (changes requested):
The tests look much much better and easier to read. Thanks for doing this. :)

src/Google/Ads/GoogleAds/Util/V1/GoogleAdsFailures.php
```diff
@@ -0,0 +1,51 @@
...
+{
+    /**
+     * Unpacks a single GoogleAdsFailure from an Any instance.
+     *
+     * @param Any an Any instance to unpack
+     * @return GoogleAdsFailure
+     */
+    public static function fromAny(Any $any)
+    {
+        return $any->unpack();
```
michaelbausor: A thought here - do you want to check that the unpacked message is actually the expected GoogleAdsFailure class, and if not, to either return null or throw an exception? Right now you could call this method on an Any object containing anything, and it will succeed and give you back that object.

src/Google/Ads/GoogleAds/Util/V0/PartialFailures.php
Raibaz: Cool, then I think it's best to merge #41 first then rebase this on top of it and remove this function and its usage in the example.

WDYT?

src/Google/Ads/GoogleAds/Util/V1/GoogleAdsFailures.php
Raibaz: Good point, especially since the clients are going to assume that the returned object is a GoogleAdsFailure. 

I think in this case throwing an exception would be the best solution, as unpacking a message that is not a GoogleAdsFailure here is really a wrong usage of this function. 

However, I don't have a strong opinion on what exception we should throw here - InvalidArgumentException? A custom exception? @fiboknacky, WDYT?

tests/Google/Ads/GoogleAds/V0/Util/GoogleAdsFailuresTest.php
Raibaz: Done.

.gitignore
Raibaz: Done, sending a separate PR for this.

src/Google/Ads/GoogleAds/Util/V0/PartialFailures.php
fiboknacky: Yep. I just merged.

src/Google/Ads/GoogleAds/Util/V1/GoogleAdsFailures.php
fiboknacky: InvalidArgumentException SGTM.

src/Google/Ads/GoogleAds/Util/V0/PartialFailures.php
Raibaz: Ok, removed this, thanks!

src/Google/Ads/GoogleAds/Util/V1/GoogleAdsFailures.php
Raibaz: Done.

composer.json
```diff
@@ -13,7 +13,7 @@
     "monolog/monolog": "^1.23.0"
   },
   "require-dev": {
-    "phpunit/phpunit": "^4.8.35 || ^5.7",
+    "phpunit/phpunit": "^7.5",
```
fiboknacky: Probably need a rebase?

examples/BasicOperations/HandlePartialFailure.php
```diff
@@ -0,0 +1,191 @@
...
+
+        $adGroupOperation2 = new AdGroupOperation();
+        $adGroupOperation2->setCreate($adGroup2);
+        $operations[] = $adGroupOperation2;
+
+        $adGroupOperation3 = new AdGroupOperation();
+        $adGroupOperation3->setCreate($adGroup3);
+        $operations[] = $adGroupOperation3;
+        
+        // Issues the mutate request, setting partialFailure=true.
```
fiboknacky: "Issues a mutate request by enabling partial failure mode."

src/Google/Ads/GoogleAds/Lib/V0/LogMessageFormatter.php
```diff
@@ -114,6 +116,14 @@ public function formatDetail(
         );
 
         if ($status->code === 0) {
+            // Initialize unused GoogleAdsFailure object when there are partial failures returned.
```
fiboknacky: Probably you need a rebase?

tests/Google/Ads/GoogleAds/V0/Util/GoogleAdsErrorsTest.php
```diff
@@ -0,0 +1,114 @@
+<?php
```
fiboknacky: Path still wrong.

fiboknacky: Please fix for all files of v0 and v1.

fiboknacky (changes requested):

tests/Google/Ads/GoogleAds/V0/Util/GoogleAdsErrorsTest.php
Raibaz: Done.

src/Google/Ads/GoogleAds/Lib/V0/LogMessageFormatter.php
Raibaz: Rebased.

composer.json
Raibaz: Rebased.

examples/BasicOperations/HandlePartialFailure.php
fiboknacky: Not done.

fiboknacky (approved):
Please fix one small unresolved nit.

PR #37, opened 2019-03-15
Added docker configuration files and docker installation explanation …

Raibaz:
…in the README

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM php:7.1-apache
+
+RUN apt-get update && apt-get install -y libxml2 zlib1g-dev git unzip
+
+RUN docker-php-ext-install bcmath
+
+RUN curl -sS https://getcomposer.org/installer | php
+RUN mv composer.phar /usr/local/bin/composer
+
+RUN pecl install grpc
+RUN echo 'extension=grpc.so' >> $PHP_INI_DIR/conf.d/grpc.ini
+
+# Create empty credentials to make sure the unit tests run successfully
+RUN echo '{"type": "authorized_user","client_id": "","client_secret": "","refresh_token": ""}' >> /root/emptycredentials.json
+ENV GOOGLE_APPLICATION_CREDENTIALS /root/emptycredentials.json
+
+WORKDIR "/google-ads-php"
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN mv composer.phar /usr/local/bin/composer
 RUN pecl install grpc
 RUN echo 'extension=grpc.so' >> $PHP_INI_DIR/conf.d/grpc.ini
 
-# Create empty credentials to make sure the unit tests run successfully
+# Create empty credentials to make sure the unit tests run successfully.
 RUN echo '{"type": "authorized_user","client_id": "","client_secret": "","refresh_token": ""}' >> /root/emptycredentials.json
 ENV GOOGLE_APPLICATION_CREDENTIALS /root/emptycredentials.json
 
--- a/README.md
+++ b/README.md
@@ -107,6 +107,19 @@ API](https://developers.google.com/google-ads/api/docs/start).
     the examples require parameters. You can see what are required by running
     code examples with `--help` as a command-line parameter.
 
+## Running in a Docker container
+
+You can also execute the examples and develop your Google Ads API application inside a Docker 
+container, built using the provided [Dockerfile](Dockerfile).
+
+To do so, provided that you have a Docker environment [installed](https://docs.docker.com/install/),
+you can run the following command in the directory where you cloned this repository:
+
+    docker-compose up --build -d
+
+This will provide a Docker container instance with all the required dependencies already installed
+and the 
+
 ## Basic usage
 
 To issue requests via the Google Ads API, you first need to create a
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -107,6 +107,19 @@ API](https://developers.google.com/google-ads/api/docs/start).
...
+You can also execute the examples and develop your Google Ads API application inside a Docker 
+container, built using the provided [Dockerfile](Dockerfile).
+
+To do so, provided that you have a Docker environment [installed](https://docs.docker.com/install/),
+you can run the following command in the directory where you cloned this repository:
+
+    docker-compose up --build -d
+
+This will provide a Docker container instance with all the required dependencies already installed
+and the
```
fiboknacky: Finish your sentence.

docker-compose.yml
```diff
@@ -0,0 +1,7 @@
+version: '2'
+services:
+  web:
+    build: .
+    volumes: 
+      - ./:/google-ads-php
+      - ~/:/root
```
fiboknacky: Can you add a new line here too?

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+RUN docker-php-ext-install bcmath
+
+RUN curl -sS https://getcomposer.org/installer | php
+RUN mv composer.phar /usr/local/bin/composer
+
+RUN pecl install grpc
+RUN echo 'extension=grpc.so' >> $PHP_INI_DIR/conf.d/grpc.ini
+
+# Create empty credentials to make sure the unit tests run successfully
```
fiboknacky: Finish the sentence with a period.

README.md
```diff
@@ -107,6 +107,19 @@ API](https://developers.google.com/google-ads/api/docs/start).
     the examples require parameters. You can see what are required by running
     code examples with `--help` as a command-line parameter.
 
+## Running in a Docker container
```
fiboknacky: Can you move this to be just before "Logging"?

README.md
Raibaz: Done, also added some further instructions on how to get a shell to the container.

Dockerfile
Raibaz: Done.

README.md
Raibaz: Done.

docker-compose.yml
Raibaz: Done.

Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM php:7.1-apache
```
fiboknacky: So, what's the requirement about PHP version here?
If a user has installed PHP via normal procedures of his/her environments, we should just run "php examples/BasicOperations/GetCampaigns.php"?

docker-compose.yml
```diff
@@ -0,0 +1,8 @@
+version: '2'
+services:
+  web:
+    build: .
+    volumes:
```
fiboknacky: No extra space needed after "volumes:"

fiboknacky (commented):
I've tried this. Overall, looks good! Some questions pending.

Dockerfile
Raibaz: The PHP version is actually specified inside the container, so we can fix it to the one we like; in this case it's 7.1 since it's the minimum required, but we can also use a more recent version.

In case a user has already installed PHP on their machines, the docker environment will not override it, but they will be able to run `php examples/BasicOperations/GetCampaigns.php` inside the container with the PHP version specified by the container itself.

They can also run it outside of the container in their local machines of course, provided they have all the dependencies installed.

docker-compose.yml
Raibaz: Done, thanks for spotting this.

Dockerfile
fiboknacky: I hit a problem first because my machine doesn't have PHP installed in a traditional way, so there was an error saying that it couldn't find php71 binary file.
Could you add a few sentences in README.md explaining about this?
It's probably should be a knowledge of using Docker, but it's more smooth to warn the users in advance.

fiboknacky (changes requested):

Dockerfile
Raibaz: When did you get the error? When running `docker-compose up` or when trying to execute the examples? Not really sure what was happening when you got the error, I may need some more info to clarify this in the README.

fiboknacky: My error was because I set some alias for the php binary itself.
So it'll be an error like "bash: xxxxx/php: No such file or directory"
I think it's a rare case, so probably we don't need to care much now.

Probably, if you know where Docker will install PHP (perhaps, /usr/local/bin/php), you may just say so briefly.

Raibaz: This is kind of weird as I already had PHP installed on my machine when I tried it and also, AFAIK, the Docker image will take care of installing the interpreter in the container on its own, regardless of the existence or not of the interpreter in the host machine.

fiboknacky: Yes, but I guess you didn't add strange alias that would hide the originally installed PHP right?

If you're sure that Docker will install its own PHP binary and will work out of the box unless someone sets the alias that would hide it, then it's fine now.
Could you finally confirm this?

Raibaz: Yes, I'm positive about it. Let's merge this, then.

Issue #39, opened 2019-03-18
Find replacement for phpunit-mock-objects


Raibaz:
When running composer install, this message is displayed:

    Package phpunit/phpunit-mock-objects is abandoned, you should avoid using it. 
    No replacement was suggested.

Another library to mock objects should be used and phpunit-mock-objects should be removed. Perhaps [Phake](https://phake.readthedocs.io/en/2.1/) would be ok.

fiboknacky:
It's required by the old PHPUnit version, which is out of our control. We should just instead install the latest one that supports PHP 7.1, which seems to be [v7.5](https://github.com/sebastianbergmann/phpunit/tree/7.5).

Could you try install that version and run the tests?

PR #40, opened 2019-03-19
Upgraded phpunit to v7

Raibaz:
This upgrades phpunit to version 7, which is the latest that still supports PHP 7.1.

The latest version of phpunit, v8, requires PHP 7.2 or 7.3.

Besides upgrading to a newer version, this also gets rid of the transitive dependency on phpunit-mock-objects that was causing the warning referred to in issue #39.

The test suite remains all green, of course.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     "monolog/monolog": "^1.23.0"
   },
   "require-dev": {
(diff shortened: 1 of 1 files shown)


fiboknacky (approved):

fiboknacky:
Please squash and merge, and also mark #39 fixed. Thanks!

PR #41, opened 2019-03-19
Make LogMessageFormatter aware of responses containing partial failures

fiboknacky:


src/Google/Ads/GoogleAds/Lib/V0/LogMessageFormatter.php
```diff
@@ -114,6 +116,13 @@ public function formatDetail(
         );
 
         if ($status->code === 0) {
+            // Initialize unused GoogleAdsFailure object when there are partial failures returned.
+            // This is necessary, as the pool needs to be aware of this class, in order to serialize
+            // the response correctly.
+            if (!is_null($response->getPartialFailureError())
```
michaelbausor: What about the case when the response object doesn't have a getPartialFailureError method? Won't this be the case for all of the `get` style RPCs, e.g. https://github.com/googleapis/googleapis/blob/master/google/ads/googleads/v1/services/ad_group_ad_service.proto#L40

fiboknacky: Fixed. Thanks for catching.

src/Google/Ads/GoogleAds/Lib/V0/LogMessageFormatter.php
```diff
@@ -114,6 +116,14 @@ public function formatDetail(
         );
 
         if ($status->code === 0) {
+            // Initialize unused GoogleAdsFailure object when there are partial failures returned.
+            // This is necessary, as the pool needs to be aware of this class, in order to serialize
+            // the response correctly.
+            if (method_exists($response, 'getPartialFailureError')
+                && !is_null($response->getPartialFailureError())
+                && !isset(self::$unusedGoogleAdsFailure)) {
```
michaelbausor: Nit: maybe check this first, because after it is initialized the first time it will always short circuit the other checks?

michaelbausor (approved):

src/Google/Ads/GoogleAds/Lib/V0/LogMessageFormatter.php
Raibaz: +1

Raibaz (approved):
Ok with a minor nit

src/Google/Ads/GoogleAds/Lib/V0/LogMessageFormatter.php
fiboknacky: Done.

PR #48, opened 2019-04-02
Better type hinting

### LINKED ISSUE #45: Incorrect typehint for `withLoginCustomerId`
In the `GoogleAdsClientBuilder`, the typehint for the `withLoginCustomerId` is `int`. However, gRPC requires a string, rather than an int. So either the typehint would need to be changed to string, or some conversion would have to be made to make sure it is a string. If you send it as an integer, gRPC will throw `Bad metadata value given`.

Raibaz:
This resolves #45 by setting the proper type hinting in the PHPdoc for `withLoginCustomerId`, and also leverages the scalar type hinting feature introduced in PHP7 and the nullable notation introduced in PHP7.1 to enforce proper types being passed to all functions in the `GoogleAdsClientBuilder` class.

### DIFF
--- a/src/Google/Ads/GoogleAds/Lib/V1/GoogleAdsClientBuilder.php
+++ b/src/Google/Ads/GoogleAds/Lib/V1/GoogleAdsClientBuilder.php
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright 2019 Google LLC
+ * Copyright 2018 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/Google/Ads/GoogleAds/Lib/V1/GoogleAdsClientBuilder.php
+++ b/src/Google/Ads/GoogleAds/Lib/V1/GoogleAdsClientBuilder.php
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright 2018 Google LLC
+ * Copyright 2019 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -60,7 +60,7 @@ final class GoogleAdsClientBuilder implements GoogleAdsBuilder
      * @throws InvalidArgumentException if the configuration file could not be
      *     found
      */
-    public function fromFile($path = null)
+    public function fromFile(?string $path)
     {
         if ($path === null) {
             $path = self::DEFAULT_CONFIGURATION_FILENAME;
@@ -97,7 +97,7 @@ final class GoogleAdsClientBuilder implements GoogleAdsBuilder
      * @param string $developerToken
      * @return self this builder
      */
-    public function withDeveloperToken($developerToken)
+    public function withDeveloperToken(string $developerToken)
     {
         $this->developerToken = $developerToken;
         return $this;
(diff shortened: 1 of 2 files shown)


src/Google/Ads/GoogleAds/Lib/V1/GoogleAdsClientBuilder.php
```diff
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright 2018 Google LLC
+ * Copyright 2019 Google LLC
```
fiboknacky: We don't modify copyright of the existing files.

src/Google/Ads/GoogleAds/Lib/V1/GoogleAdsClientBuilder.php
```diff
@@ -112,10 +112,10 @@ public function withDeveloperToken($developerToken)
      * create a separate GoogleAdsClient instance for each manager account. Use this method to
      * set each login customer ID and call build() to create a separate instance.
      *
-     * @param int $loginCustomerId the login customer ID
+     * @param string|null $loginCustomerId the login customer ID
      * @return self this builder
      */
-    public function withLoginCustomerId($loginCustomerId)
+    public function withLoginCustomerId(?string $loginCustomerId)
```
fiboknacky: Let be consistent with the above too.
If you want to use ?string, please modify the above V0 to be the same.

fiboknacky (changes requested):

src/Google/Ads/GoogleAds/Lib/V1/GoogleAdsClientBuilder.php
Raibaz: Done.

src/Google/Ads/GoogleAds/Lib/V1/GoogleAdsClientBuilder.php
Raibaz: What do you mean? This change has been made on both the V0 and V1 classes.

fiboknacky: Sorry, I meant "fromFile()" method.

Raibaz: I think they are two different cases because fromFile can be called with no arguments (as it is called in the examples) and in that case it sets a default value. 

In my opinion, withLoginCustomerId shouldn't be called with no arguments because we don't provide a default value, so it's correct to have users call it explicitly passing null if they want to set it to null.

To clarify, calling `fromFile()` is acceptable in my opinion, while calling `withLoginCustomerId()` isn't, because we don't set a default value and if users want to unset it they should be explicitly be calling `withLoginCustomerId(null)`.

WDYT?

fiboknacky: Got it. This makes sense to me, as we don't expect users to call fromFile(null), so having the signature as fromFile(?string $fileName = null) wouldn't make sense too.

fiboknacky (approved):
Probably the test cases wouldn't be affected.
But just in case, please double check if we need changes in any test cases.

Raibaz:
The test cases were exactly where I learned that you can call withCustomerId(null) :)

So yes, they are not affected and they don't need any changes.

Issue #63, opened 2019-04-26
Example Issue AddKeywordPlan.php 


chiragvels:
Hi,

Argument 3 passed to Google\Ads\GoogleAds\V1\Services\Gapic\KeywordPlanServiceGapicClient::mutateKeywordPlans() must be of the type array, boolean given

```
$response = $keywordPlanServiceClient->mutateKeywordPlans(
            $customerId,
            [$keywordPlanOperation],
            false,
            false
        );

@param string                 $customerId   The ID of the customer whose keyword plans are being modified.
@param KeywordPlanOperation[] $operations   The list of operations to perform on individual keyword plans.
@param array                  $optionalArgs {
                                             Optional.

mutateKeywordPlans($customerId, $operations, array $optionalArgs = [])
```

Same way mutateKeywordPlanCampaigns, mutateKeywordPlanAdGroups, mutateKeywordPlanKeywords, mutateKeywordPlanNegativeKeywords has same boolean given.

2) I can see that there is attribute difference in Old and New API.
I have post this question in Forum as It might not be right to ask here.
But still I have posted, If any help from lib creator can do?

I am comparing data TargetingIdeaService.get with RequestType=STATS with KeywordPlanService.GenerateHistoricalMetrics
  
I am not getting all of the Attributes As Adwords API
For e.g. CATEGORY_PRODUCTS_AND_SERVICES, EXTRACTED_FROM_WEBPAGE, AVERAGE_CPC

Also there is not matching with new UI like attaributes “Top of page bid (low range)”, “Top of page bid (high range)”, Competition (indexed value), Organic average position (“Organic avg. position”), Organic impression share (“Organic impr. share”) 

Thanks,

Raibaz:
Thanks for noticing this, the PR I just merged should fix it.

PR #64, opened 2019-04-30
Fixed partial failure and validate only parameters in mutate calls

Raibaz:
This fixes issue #63 by updating the mutate calls to the proper format.


jradcliff:
Small suggestion: when merging, please update the commit comment to mention keyword plans so it's clear to readers of commit history which example you changed. The current wording could give the impression that this impacts *all* mutate calls.

PR #75, opened 2019-05-13
Use enum getters in AM and CM examples

PierrickVoulet:
Use enum getters to display value names instead of IDs in the Account Management and CampaignManagement code examples.

### DIFF
--- a/examples/AccountManagement/GetAccountChanges.php
+++ b/examples/AccountManagement/GetAccountChanges.php
@@ -121,7 +121,7 @@ class GetAccountChanges
             /** @var GoogleAdsRow $googleAdsRow */
             printf(
                 "On %s, change status '%s' shows resource '%s' with type '%s' and status '%s'.%s",
-                $googleAdsRow->getChangeStatus()->getLastChangeDateTime()->getValue(),
+                $googleAdsRow->getChangeStatus()->getLastChangeDateTimeValue(),
                 $googleAdsRow->getChangeStatus()->getResourceName(),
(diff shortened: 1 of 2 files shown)


examples/AccountManagement/GetAccountChanges.php
```diff
@@ -118,18 +119,13 @@ public static function runExample(GoogleAdsClient $googleAdsClient, $customerId)
...
             /** @var GoogleAdsRow $googleAdsRow */
-            // Note that the resource type and resource status printed below are enum values.
-            // For example, a value of 3 will be returned when the resource type is 'AD_GROUP'.
-            // A mapping of enum names to values can be found in the following files:
-            // * ChangeStatusResourceType.php (for resource type)
-            // * ChangeStatusOperation.php (for resource status)
             printf(
-                "On %s, change status '%s' shows resource '%s' with type %d and status %d.%s",
+                "On %s, change status '%s' shows resource '%s' with type '%s' and status '%s'.%s",
                 $googleAdsRow->getChangeStatus()->getLastChangeDateTime()->getValue(),
```
Raibaz: Should we also change likes like this to `$googleAdsRow->getChangeStatus()->getLastChangeDateTimeValue()`?

@fiboknacky, WDYT?

fiboknacky: I thought Pierrick would do that as a part of other PRs.
But yes, please do it here as it's still somehow related changes and it saves number of PRs.

PierrickVoulet: Done

PR #82, opened 2019-05-17
Added the dynamic feed example

Raibaz:
Ported from https://github.com/googleads/google-ads-java/pull/91/files

PR #83, opened 2019-05-20
Add the AddGmailAd code example

PierrickVoulet:


examples/AdvancedOperations/AddGmailAd.php
```diff
@@ -0,0 +1,251 @@
...
+
+        // Returns the created media file resource names.
+        return [
+            'logoResourceName' => $response->getResults()[0]->getResourceName(),
+            'marketingImageResourceName' => $response->getResults()[1]->getResourceName()
+        ];
+    }
+
+    /**
+     * Adds the gmail ad.
```
Raibaz: Gmail should be capitalized as it is in other comments.

examples/AdvancedOperations/AddGmailAd.php
```diff
@@ -0,0 +1,251 @@
...
+            'ad_group' => $adGroupResourceName
+        ]);
+
+        // Creates an ad group ad operation.
+        $adGroupAdOperation = new AdGroupAdOperation();
+        $adGroupAdOperation->setCreate($adGroupAd);
+
+        // Issues a mutate request to add the ad group ad.
+        $adGroupAdServiceClient = $googleAdsClient->getAdGroupAdServiceClient();
+        $response = $adGroupAdServiceClient->mutateAdGroupAds($customerId, [$adGroupAdOperation]);
```
Raibaz: Should we explicitly handle the cases where the provided campaignId and/or adGroupId don't allow the creation of a GmailAd?

PierrickVoulet: The prerequisite is explained in the description of the class and I could not find any other code example doing such handling so I do not think it is required here.

But I agree that it could be interesting to demonstrate how to handle errors like this one, may be a separated code example in the "ErrorHandling" session would be a better fit?

examples/AdvancedOperations/AddGmailAd.php
PierrickVoulet: Done

examples/AdvancedOperations/AddGmailAd.php
Raibaz: The add dynamic feed example does something similar [here](https://github.com/googleads/google-ads-php/pull/82/files#diff-7a17d89c2c186ebdd6a49a2eb5280749R411)

PierrickVoulet: Would a test run at the beginning of the example that explicitly makes a request to retrieve the ad group information and throw an exception if invalid be good?

Raibaz: Mmm not sure about that as it would probably add some code that is not really relevant to the purpose of the example. I'd rather just let the thrown exception bubble up and break the example execution in case the campaign or ad group don't support GmailAd.

PierrickVoulet: I tried with a bad campaign type and an exception already bubbles up with a "OPERATION_NOT_PERMITTED_FOR_CAMPAIGN_TYPE" error code and some logs:
-----
Request with ID 'XXX' has failed.
Google Ads failure details:
	operation_access_denied_error: This operation is not permitted on this campaign type
-----

This seems pretty adapted so I do not think that catching/parsing the exception and then throwing a new one would add anything to this example?

Raibaz: I agree, let's leave it as it is.

examples/AdvancedOperations/AddGmailAd.php
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    /**
+     * Adds the image files.
+     *
+     * @param GoogleAdsClient $googleAdsClient the Google Ads API client
+     * @param int $customerId the client customer ID
+     * @return array the image file resource names
+     */
+    private static function addMediaFiles(
```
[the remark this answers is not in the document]
PierrickVoulet: Done

examples/AdvancedOperations/AddGmailAd.php
```diff
@@ -0,0 +1,251 @@
...
+        GoogleAdsClient $googleAdsClient,
+        int $customerId,
+        int $adGroupId
+    ) {
+        $mediaFiles = self::addMediaFiles($googleAdsClient, $customerId);
+        self::addGmailAd($googleAdsClient, $customerId, $adGroupId, $mediaFiles);
+    }
+
+    /**
+     * Adds the image files.
```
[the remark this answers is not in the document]
PierrickVoulet: Done

examples/AdvancedOperations/AddGmailAd.php
```diff
@@ -121,11 +121,11 @@ public static function runExample(
     }
 
     /**
-     * Adds the image files.
+     * Adds the media files based on URLs.
```
fiboknacky: Nit: "... media files by using the class constants."

PierrickVoulet: Done.

PR #87, opened 2019-05-21
Add the LinkManagerToClient code example

PierrickVoulet:


examples/AccountManagement/LinkManagerToClient.php
```diff
@@ -0,0 +1,259 @@
...
+use Google\Ads\GoogleAds\V1\Errors\GoogleAdsError;
+use Google\Ads\GoogleAds\V1\Resources\CustomerClientLink;
+use Google\Ads\GoogleAds\V1\Resources\CustomerManagerLink;
+use Google\Ads\GoogleAds\V1\Services\CustomerClientLinkOperation;
+use Google\Ads\GoogleAds\V1\Services\CustomerManagerLinkOperation;
+use Google\ApiCore\ApiException;
+use Google\Protobuf\StringValue;
+
+/**
+ * This example demonstrates how to link an existing Google Ads manager customer
```
Raibaz: Nit: wouldn't it be better to say "manager customer account" and "client customer account"?

examples/AccountManagement/LinkManagerToClient.php
```diff
@@ -0,0 +1,259 @@
...
+        printf(
+            "An invitation has been extended from the manager customer %d" .
+            " to the client customer %d with the customer client link resource name '%s'.%s",
+            $managerCustomerId,
+            $clientCustomerId,
+            $customerClientLinkResourceName,
+            PHP_EOL
+        );
+
+        // Creates a query that retrieves the manager link ID of the customer client link
```
Raibaz: In all examples, when we do a mutate operation and then a query to retrieve the result, the query is in a separate function. Can you please extract this to a separate function as well? Not just for coherence with the other examples, but also because you now have a 68-line function, which looks pretty long to me.

examples/AccountManagement/LinkManagerToClient.php
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    /**
+     * Creates a Google Ads client based on the default configuration file
+     * and a given login customer id.
+     *
+     * @param int $loginCustomerId  the login customer ID
+     * @return GoogleAdsClient the created client
+     */
+    private static function createGoogleClient(int $loginCustomerId)
```
fiboknacky: Should be "createGoogleAdsClient"

examples/AccountManagement/LinkManagerToClient.php
fiboknacky: +1

PR #88, opened 2019-05-22
Changes for release v1_3.

Raibaz:


examples/BasicOperations/GetCampaigns.php
```diff
@@ -98,7 +98,7 @@ public static function runExample(GoogleAdsClient $googleAdsClient, $customerId)
         // Issues a search request by specifying page size.
         $response =
             $googleAdsServiceClient->search($customerId, $query, ['pageSize' => self::PAGE_SIZE]);
-
+
```
PierrickVoulet: Nit: extra spaces

Raibaz: Fixed, thanks.

fiboknacky: It's not fixed yet. There should be no spaces in this line at all. Should be just a blank line, like it's never modified.

Raibaz:
@fiboknacky the GetCampaigns example is now reverted to how it was before the release, you can see it since it doesn't appear in https://github.com/googleads/google-ads-php/pull/88/files

PR #90, opened 2019-05-23
Add AddCompleteCampaignsUsingMutateJob example

fiboknacky:
Also added + modified some necessary classes.

### DIFF
--- a/examples/CampaignManagement/AddCompleteCampaignsUsingMutateJob.php
+++ b/examples/CampaignManagement/AddCompleteCampaignsUsingMutateJob.php
@@ -1,0 +1,483 @@
+<?php
+/**
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Ads\GoogleAds\Examples\CampaignManagement;
+
+require __DIR__ . '/../../vendor/autoload.php';
+
+use GetOpt\GetOpt;
+use Google\Ads\GoogleAds\Examples\Utils\ArgumentNames;
+use Google\Ads\GoogleAds\Examples\Utils\ArgumentParser;
+use Google\Ads\GoogleAds\Lib\OAuth2TokenBuilder;
+use Google\Ads\GoogleAds\Lib\V1\GoogleAdsClient;
+use Google\Ads\GoogleAds\Lib\V1\GoogleAdsClientBuilder;
+use Google\Ads\GoogleAds\Lib\V1\GoogleAdsException;
+use Google\Ads\GoogleAds\Util\V1\ResourceNames;
+use Google\Ads\GoogleAds\V1\Common\ExpandedTextAdInfo;
+use Google\Ads\GoogleAds\V1\Common\KeywordInfo;
+use Google\Ads\GoogleAds\V1\Common\ManualCpc;
+use Google\Ads\GoogleAds\V1\Enums\AdGroupAdStatusEnum\AdGroupAdStatus;
+use Google\Ads\GoogleAds\V1\Enums\AdGroupCriterionStatusEnum\AdGroupCriterionStatus;
+use Google\Ads\GoogleAds\V1\Enums\AdGroupTypeEnum\AdGroupType;
+use Google\Ads\GoogleAds\V1\Enums\AdvertisingChannelTypeEnum\AdvertisingChannelType;
+use Google\Ads\GoogleAds\V1\Enums\BudgetDeliveryMethodEnum\BudgetDeliveryMethod;
+use Google\Ads\GoogleAds\V1\Enums\CampaignStatusEnum\CampaignStatus;
+use Google\Ads\GoogleAds\V1\Enums\KeywordMatchTypeEnum\KeywordMatchType;
+use Google\Ads\GoogleAds\V1\Errors\GoogleAdsError;
+use Google\Ads\GoogleAds\V1\Errors\GoogleAdsFailure;
+use Google\Ads\GoogleAds\V1\Resources\Ad;
+use Google\Ads\GoogleAds\V1\Resources\AdGroup;
+use Google\Ads\GoogleAds\V1\Resources\AdGroupAd;
+use Google\Ads\GoogleAds\V1\Resources\AdGroupCriterion;
+use Google\Ads\GoogleAds\V1\Resources\Campaign;
+use Google\Ads\GoogleAds\V1\Resources\CampaignBudget;
+use Google\Ads\GoogleAds\V1\Resources\CampaignCriterion;
+use Google\Ads\GoogleAds\V1\Services\AdGroupAdOperation;
+use Google\Ads\GoogleAds\V1\Services\AdGroupCriterionOperation;
+use Google\Ads\GoogleAds\V1\Services\AdGroupOperation;
+use Google\Ads\GoogleAds\V1\Services\CampaignBudgetOperation;
+use Google\Ads\GoogleAds\V1\Services\CampaignCriterionOperation;
+use Google\Ads\GoogleAds\V1\Services\CampaignOperation;
+use Google\Ads\GoogleAds\V1\Services\CampaignServiceClient;
+use Google\Ads\GoogleAds\V1\Services\MutateJobResult;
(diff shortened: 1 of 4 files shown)


examples/CampaignManagement/AddCompleteCampaignsUsingMutateJob.php
```diff
@@ -0,0 +1,486 @@
...
+    const NUMBER_OF_CAMPAIGNS_TO_ADD = 2;
+    const NUMBER_OF_ADGROUPS_TO_ADD = 2;
+    const NUMBER_OF_KEYWORDS_TO_ADD = 5;
+    const POLL_FREQUENCY_SECONDS = 1;
+    const MAX_TOTAL_POLL_INTERVAL_SECONDS = 60;
+
+    const PAGE_SIZE = 1000;
+
+    /** @var int the negative temporary ID used in mutate job operations. */
+    private static $temporaryId = 0;
```
Raibaz: It's a bit unclear here IMHO: if it's negative, why are you initializing it to 0?

examples/CampaignManagement/AddCompleteCampaignsUsingMutateJob.php
```diff
@@ -0,0 +1,486 @@
...
+     * @param GoogleAdsClient $googleAdsClient the Google Ads API client
+     * @param int $customerId the client customer ID without hyphens
+     */
+    public static function runExample(GoogleAdsClient $googleAdsClient, $customerId)
+    {
+        $mutateJobServiceClient = $googleAdsClient->getMutateJobServiceClient();
+
+        // Creates a new mutate job.
+        $mutateJobResourceName =
+            $mutateJobServiceClient->createMutateJob($customerId)->getResourceName();
```
Raibaz: In other examples we return the resource name directly rather than returning an object and then calling getResourceName externally, do you think it makes sense to change this for coherence?

examples/CampaignManagement/AddCompleteCampaignsUsingMutateJob.php
```diff
@@ -0,0 +1,486 @@
...
+     * corresponding mutate operations.
+     *
+     * @param int $customerId the client customer ID
+     * @return MutateOperation[] the mutate operations to be added to a mutate job
+     */
+    private static function buildAllOperations(int $customerId)
+    {
+        $mutateOperations = [];
+
+        // Creates a new campaign budget operation and add it to the array of mutate operations.
```
Raibaz: Nit "and adds it"

examples/CampaignManagement/AddCompleteCampaignsUsingMutateJob.php
```diff
@@ -0,0 +1,486 @@
...
+        $campaignBudgetOperation = self::buildCampaignBudgetOperation($customerId);
+        $mutateOperations[] =
+            new MutateOperation(['campaign_budget_operation' => $campaignBudgetOperation]);
+
+        // Creates new campaign operations and add them to the array of mutate operations.
```
Raibaz: Nit: "and adds them"

examples/CampaignManagement/AddCompleteCampaignsUsingMutateJob.php
```diff
@@ -0,0 +1,486 @@
...
+            $campaignBudgetOperation->getCreate()->getResourceName()
+        );
+        $mutateOperations = array_merge($mutateOperations, array_map(
+            function (CampaignOperation $campaignOperation) {
+                return new MutateOperation(['campaign_operation' => $campaignOperation]);
+            },
+            $campaignOperations
+        ));
+
+        // Creates new campaign criterion operations and add them to the array of mutate operations.
```
Raibaz: Nit: "and adds them"

examples/CampaignManagement/AddCompleteCampaignsUsingMutateJob.php
```diff
@@ -0,0 +1,486 @@
...
+        $mutateOperations = array_merge($mutateOperations, array_map(
+            function (CampaignCriterionOperation $campaignCriterionOperation) {
+                return new MutateOperation(
+                    ['campaign_criterion_operation' => $campaignCriterionOperation]
+                );
+            },
+            self::buildCampaignCriterionOperations($campaignOperations)
+        ));
+
+        // Creates new campaign ad group operations and add them to the array of mutate operations.
```
Raibaz: Nit: "and adds them"

examples/CampaignManagement/AddCompleteCampaignsUsingMutateJob.php
```diff
@@ -0,0 +1,486 @@
...
+        $adGroupOperations = self::buildAdGroupOperations($customerId, $campaignOperations);
+        $mutateOperations = array_merge($mutateOperations, array_map(
+            function (AdGroupOperation $adGroupOperation) {
+                return new MutateOperation(['ad_group_operation' => $adGroupOperation]);
+            },
+            self::buildAdGroupOperations($customerId, $campaignOperations)
+        ));
+
+        // Creates new ad group criterion operations and add them to the array of mutate operations.
```
Raibaz: Nit: "and adds them"

examples/CampaignManagement/AddCompleteCampaignsUsingMutateJob.php
```diff
@@ -0,0 +1,486 @@
...
+            self::buildCampaignCriterionOperations($campaignOperations)
+        ));
+
+        // Creates new campaign ad group operations and add them to the array of mutate operations.
+        $adGroupOperations = self::buildAdGroupOperations($customerId, $campaignOperations);
+        $mutateOperations = array_merge($mutateOperations, array_map(
+            function (AdGroupOperation $adGroupOperation) {
+                return new MutateOperation(['ad_group_operation' => $adGroupOperation]);
+            },
+            self::buildAdGroupOperations($customerId, $campaignOperations)
```
Raibaz: Why are you calling buildAdGroupOperations twice? I think you should either keep the inlined call inside array_map or the one before array_merge, not both.

Raibaz (changes requested):

examples/CampaignManagement/AddCompleteCampaignsUsingMutateJob.php
fiboknacky: I changed to start from -1 instead.

examples/CampaignManagement/AddCompleteCampaignsUsingMutateJob.php
fiboknacky: This is probably different situation.
In other cases, we didn't get "resource names", but get other attributes coming with the object (like getFeed()).
This case is like other examples such as AddCampaign, in which we just get a resource name returned directly from the server.

examples/CampaignManagement/AddCompleteCampaignsUsingMutateJob.php
fiboknacky: This is a mistake. Thanks for catching.
I want to optimize doing buildAdGroupOperations but forgot to remove another occurrences.

Raibaz (approved):

Raibaz:
Ok pending a couple of nits int he comments

michaelbausor (approved):

examples/CampaignManagement/AddCompleteCampaignsUsingMutateJob.php
```diff
@@ -0,0 +1,483 @@
...
+            "Mutate job with resource name '%s' has been executed.%s",
+            $mutateJobResourceName,
+            PHP_EOL
+        );
+
+        // Polls the server until the mutate job execution finishes by setting the initial poll
+        // delay time and the total time to wait before time-out.
+        $operationResponse->pollUntilComplete([
+            'initialPollDelayMillis' => self::POLL_FREQUENCY_SECONDS * 1000,
+            'totalPollTimeoutMillis' => self::MAX_TOTAL_POLL_INTERVAL_SECONDS * 1000
```
AnashOommen: This is fine for now, but talk to Ben / Bob about whether we can autogenerate reasonable defaults for these parameters.

examples/CampaignManagement/AddCompleteCampaignsUsingMutateJob.php
```diff
@@ -0,0 +1,483 @@
...
+     */
+    private static function buildCampaignBudgetOperation(int $customerId)
+    {
+        // Creates a campaign budget operation.
+        return new CampaignBudgetOperation([
+            'create' => new CampaignBudget([
+                // Creates a resource name using the temporary ID.
+                'resource_name' => ResourceNames::forCampaignBudget(
+                    $customerId,
+                    self::$temporaryId--
```
AnashOommen: Can we extract the temporaryId generation to a small method like getNextTempId()? It's easier to read that way and understand what's happening.

AnashOommen (approved):
I left some comments, but since they are minor, I am marking as LGTM. Pls fix before submitting.

examples/CampaignManagement/AddCompleteCampaignsUsingMutateJob.php
fiboknacky: Sure.

Issue #98, opened 2019-05-28
ResourceNames class missing entities


projectyang:
Noticed [ResourceNames.php](https://github.com/googleads/google-ads-php/blob/master/src/Google/Ads/GoogleAds/Util/V1/ResourceNames.php) is currently missing entities. (e.g. ExtensionFeedItemServiceClient)

Happy to volunteer and finish adding the rest.

Raibaz:
Hi, thanks for spotting this: yes we are aware that the ResourceNames class is far from complete and it lacks functions for several resources, so if you could send a PR adding the ones you need we'd really appreciate it :)

Raibaz:
Fixed by #100.

PR #100, opened 2019-05-29
googleads/google-ads-php#98 Added more resource names

MaksTech:


src/Google/Ads/GoogleAds/Util/V1/ResourceNames.php
```diff
@@ -17,17 +17,57 @@
...
 
+use Google\Ads\GoogleAds\V1\Services\AccountBudgetServiceClient;
+use Google\Ads\GoogleAds\V1\Services\AccountBudgetProposalServiceClient;
+use Google\Ads\GoogleAds\V1\Services\AdGroupServiceClient;
 use Google\Ads\GoogleAds\V1\Services\AdGroupAdServiceClient;
+use Google\Ads\GoogleAds\V1\Services\AdGroupAdLabelServiceClient;
+use Google\Ads\GoogleAds\V1\Services\AdGroupAudienceViewServiceClient;
 use Google\Ads\GoogleAds\V1\Services\AdGroupBidModifierServiceClient;
 use Google\Ads\GoogleAds\V1\Services\AdGroupCriterionServiceClient;
-use Google\Ads\GoogleAds\V1\Services\AdGroupServiceClient;
```
Raibaz: Please preserve the alphabetic sorting of uses.

src/Google/Ads/GoogleAds/Util/V1/ResourceNames.php
```diff
@@ -60,9 +124,55 @@ public static function forAdGroup($customerId, $adGroupId)
      * @param int $adId the ad ID
      * @return string the ad group ad resource name
      */
-    public static function forAdGroupAd($customerId, $adGroupId, $adId)
-    {
-        return AdGroupAdServiceClient::adGroupAdName($customerId, "{$adGroupId}~{$adId}");
+    public static function forAdGroupAd(
```
Raibaz: These can fit on a single line.

src/Google/Ads/GoogleAds/Util/V1/ResourceNames.php
```diff
@@ -89,14 +202,218 @@ public static function forAdGroupBidModifier($customerId, $adGroupId, $criterion
      * @param int $criterionId the criterion ID
      * @return string the ad group criterion resource name
      */
-    public static function forAdGroupCriterion($customerId, $adGroupId, $criterionId)
-    {
+    public static function forAdGroupCriterion(
```
Raibaz: These can fit on a single line.

src/Google/Ads/GoogleAds/Util/V1/ResourceNames.php
```diff
@@ -89,14 +202,218 @@ public static function forAdGroupBidModifier($customerId, $adGroupId, $criterion
...
+
+    /**
+     * Generates resource name for an ad group feed.
+     *
+     * @param int $customerId the customer ID
+     * @param int $adGroupId the ad group ID
+     * @param int $feedId the feed Id
+     * @return string the ad group feed resource name
+     */
+    public static function forAdGroupFeed(
```
Raibaz: These can fit on a single line.

MaksTech:
Are methods also supposed to be in the same order as the use clauses? So in alphabetic order too

Raibaz:
Yep having the methods in alphabetic order is helpful to find them quickly, thanks!

MaksTech:
Okay, I'm also going to add the rest of the resource names

MaksTech:
Just noticed, [PaymentsAccount](https://github.com/googleads/google-ads-php/blob/master/src/Google/Ads/GoogleAds/V1/Services/Gapic/PaymentsAccountServiceGapicClient.php) doesn't have template, name and parse name methods. Should I add them?

Raibaz:
No, please don't add them: the `PaymentsAccount` class is actually generated by the proto compiler, so any manual change would be lost in the next release. 

Besides, it's intentional for it not to have the template, name and parse methods as the PaymentsAccount resource cannot be queried by itself but only as part of a billing setup.

src/Google/Ads/GoogleAds/Util/V1/ResourceNames.php
```diff
@@ -162,88 +504,1006 @@ public static function forCampaignCriterion($customerId, $campaignId, $criterion
...
+    public static function forParentalStatusView($customerId, $adGroupId, $criterionId)
+    {
+        return ParentalStatusViewServiceClient::parentalStatusViewName(
+            $customerId,
+            "{$adGroupId}~{$criterionId}"
+        );
+    }
+
+    /**
+     * Generates resource name for a paid organic search term view.
```
fiboknacky: Please fix wrong PHPDoc here and in @return

MaksTech: Done

PR #106, opened 2019-06-04
Add AddCampaignDraft code example

PierrickVoulet:


examples/CampaignManagement/AddCampaignDraft.php
```diff
@@ -0,0 +1,134 @@
...
+use Google\Ads\GoogleAds\Util\V1\ResourceNames;
+use Google\Ads\GoogleAds\V1\Errors\GoogleAdsError;
+use Google\Ads\GoogleAds\V1\Resources\CampaignDraft;
+use Google\Ads\GoogleAds\V1\Services\CampaignDraftOperation;
+use Google\ApiCore\ApiException;
+use Google\Protobuf\StringValue;
+
+/**
+ * This example adds a campaign draft for a campaign. Make sure you specify a
+ * campaign that has a budget with explicitly_shared set to false.
```
fiboknacky: "budget with explicitly_shared set to false" is a bit harder to process.
Can we say "non-shared budget" instead?

examples/CampaignManagement/AddCampaignDraft.php
```diff
@@ -0,0 +1,134 @@
...
+ */
+class AddCampaignDraft
+{
+    const CUSTOMER_ID = 'INSERT_CUSTOMER_ID_HERE';
+    const CAMPAIGN_ID = 'INSERT_CAMPAIGN_ID_HERE';
```
fiboknacky: In order to avoid problems in the future, I'd recommend creating another constant as "BASE_CAMPAIGN_ID" for this use.
And also modify the name of the parameter in line 105 to use $baseCampaignId

examples/CampaignManagement/AddCampaignDraft.php
```diff
@@ -0,0 +1,134 @@
...
+
+    public static function main()
+    {
+        // Either pass the required parameters for this example on the command line, or insert them
+        // into the constants above.
+        $options = (new ArgumentParser())->parseCommandArguments([
+            ArgumentNames::CUSTOMER_ID => GetOpt::REQUIRED_ARGUMENT,
+            ArgumentNames::CAMPAIGN_ID=> GetOpt::REQUIRED_ARGUMENT
```
Raibaz: Nit: space before the  =>

examples/CampaignManagement/AddCampaignDraft.php
PierrickVoulet: Done

examples/CampaignManagement/AddCampaignDraft.php
PierrickVoulet: Done

examples/CampaignManagement/AddCampaignDraft.php
PierrickVoulet: Done

PR #109, opened 2019-06-05
Remove unnecessary dependencies

PierrickVoulet:
The main purpose is to clarify the installation instructions and remove unnecessary dependencies for better stability and maintenance.

PR #111, opened 2019-06-06
Remove mocking of protobuf classes

michaelbausor:
The protobuf c extension does not work with mocks - it causes a segmentation fault. See https://github.com/protocolbuffers/protobuf/issues/4107.

This PR replaces mocks with actual instances of protobuf classes.

cc @AnashOommen


Raibaz:
@PierrickVoulet do we need this if we remove the dependency from the C extension?

PierrickVoulet:
@Raibaz yes we definitely need this because it is a blocker for performance benchmarking. #109 should only be considered as a temporary mitigation before we take a long-term decision regarding what Protobuf implementation(s) we want to support/recommend based on potential performance benefit but not only.

Issue #60, opened 2019-04-24
Service Client Config >> Timeout Milles Method Search


jeaaas:
Hi,

I've an error using the calls for the Reporting, using the default parameters of the service client for the method Search.

When I'm using the default **Service Client Config** with the **timeout_millis** to 3600000 it's returning an exception.

The resource that I'm ref. to it's in the json file: google-ads-php\src\Google\Ads\GoogleAds\V1\Services\resources\**google_ads_service_client_config.json**

Setting the timeout_millis param to 600000 it's working.

Please, Could you look into this?

Thank you in advance,
Best,
Jessica

Raibaz:
Hi, is this happening with some specific GAQL queries? Can you please send some examples of queries that run into this error?

Also, what exception are you getting?

And also, have you tried setting the `page_size` parameter in your calls to the search function? That may help in reducing the response time of your queries and avoid hitting the timeout.

jeaaas:
Hi,

I tried using the query for the campaign: "SELECT campaign.id, metrics.impressions, metrics.clicks, metrics.cost_micros, metrics.average_position 
FROM campaign 
WHERE segments.date BETWEEN '2019-04-29' AND '2019-04-29' 
ORDER BY metrics.impressions DESC", setting the page size to 1000.

The exception that returns in this call with the timeout to 360000 is "Timeval expects a long" in the file "grpc\grpc\src\lib\AbstractCall.php" in line 55.

Raibaz:
Not sure which part of the configuration you are changing here; are you referring to this?

    "methods": {
        "Search": {
          "timeout_millis": 3600000,
          "retry_codes_name": "non_idempotent",
          "retry_params_name": "default"
        },
        "Mutate": {
          "timeout_millis": 600000,
          "retry_codes_name": "non_idempotent",
          "retry_params_name": "default"
        }
      }

I have been changing both the timeout_millis values here with the values you mentioned but haven't been able to reproduce the error  you are getting.

PaoloNardini:
Hi, I'm a Jessica's collegue experiencing the same problem. I can confirm that setting Search timeout_millis to 3,600,000 (default value) raise the exception "Timeval expects a long".
It could be a problem related to Windows X86 version of grpc implementation. 
Grpc Timeval class expects a long value in constructor, while the timeout value is passed as a php int. I'm not sure how to workaround if this is the problem.

PaoloNardini:
[...]

PR #116, opened 2019-06-11
Update service-level timeout and method-level idempotency configuration

fiboknacky:
This change updates our client-side timeout settings to more closely match those that are defined by the server in order to prevent the client library from ending requests too soon and ultimately reducing the amount of DEADLINE_EXCEEDED errors that are generated.

Before this change all services were configured with a 20 second timeout in the client library. The majority of services have a server-side timeout of 60 seconds, and in the case of GoogleAdsService.Search there's a 1 hour server-side timeout. This change sets the client-side timeouts to the upper bound (1 hour) in order to prevent the client library from ending requests prematurely.

Additionally a few service methods (including GoogleAdsService.Search) were previously incorrectly configured to be non-idempotent. This means that such methods would not retry requests at all, causing a behavior where a search request would wait 20 seconds for a response and fail instead of retrying. With the change search requests (and a few others) will properly retry if they encounter a DEADLINE_EXCEEDED or UNAVAILABLE error from the server

### DIFF
--- a/src/Google/Ads/GoogleAds/V1/Services/resources/account_budget_proposal_service_client_config.json
+++ b/src/Google/Ads/GoogleAds/V1/Services/resources/account_budget_proposal_service_client_config.json
@@ -10,13 +10,13 @@
       },
       "retry_params": {
         "default": {
-          "initial_retry_delay_millis": 100,
+          "initial_retry_delay_millis": 5000,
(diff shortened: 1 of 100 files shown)


AnashOommen (approved):

Raibaz (approved):

Raibaz:
This is also likely going to fix #60

PR #120, opened 2019-06-13
Fix logging level issues

Raibaz:
Log level usage from the configuration was buggy, as it didn't work as mentioned in the README; request summaries and details should always be logged at the same level (INFO/WARNING for summaries and DEBUG/NOTICE for details)

This fixes the issue by making sure that the log level set in the configuration is correctly used as a filter; incidentally, it was already like this in the StreamHandler initialization but then the GoogleAdsUnaryCallLogger was always logging messages at the level set in the configuration.

It also makes some additional improvements:
- GoogleAdsLoggerTrait was now being used in a single place, so it didn't make sense for it to exist anymore, it has been moved inside GoogleAdsUnaryCallLogger
- GoogleAdsUnaryCallLogger also makes sure that the messages should be logged (i.e. their level is supported) before constructing them, preventing message formatting from being performed when it's not actually needed.


src/Google/Ads/GoogleAds/Lib/V1/GoogleAdsUnaryCallLogger.php
```diff
@@ -27,13 +28,14 @@
 final class GoogleAdsUnaryCallLogger
 {
     use ArrayTrait;
-    use GoogleAdsLoggerTrait;
 
     private $logger;
     private $logLevel;
```
fiboknacky: Could you improve the name of this variable as well, e.g., filterLevel?

src/Google/Ads/GoogleAds/Lib/V1/GoogleAdsUnaryCallLogger.php
```diff
@@ -27,13 +28,14 @@
...
     private $endpoint;
     private $logMessageFormatter;
     private $context;
+    private static $logLevelList;
```
fiboknacky: Could you use "private const" and change this to CAPITAL LETTERs instead?
It was not possible in PHP 5, but it should be fine now.

src/Google/Ads/GoogleAds/Lib/V1/GoogleAdsUnaryCallLogger.php
Raibaz: Done.

src/Google/Ads/GoogleAds/Lib/V1/GoogleAdsUnaryCallLogger.php
Raibaz: This is not possible even in PHP7.1, because const fields cannot be initialized with an expression, they have to be initialized with a literal that can be evaluated at compile time.

Thus, initializing them as we do here with `array_keys(Logger::getLevels())` will throw a fatal error.

PR #124, opened 2019-06-14
Added support for service accounts and impersonation.

fiboknacky:


src/Google/Ads/GoogleAds/Lib/OAuth2TokenBuilder.php
```diff
@@ -112,6 +120,45 @@ public function withRefreshToken($refreshToken)
...
+    }
+
+    /**
+     * Includes an email of account to impersonate when using service account
+     * flow. Optional and only applicable when using service account flow.
+     *
+     * @param string|null $impersonatedEmail
+     * @return OAuth2TokenBuilder this builder
+     */
+    public function withImpersonatedEmail(?string $impersonatedEmail) : self
```
Raibaz: Yay! PHP7 type hinting!

src/Google/Ads/GoogleAds/Lib/OAuth2TokenBuilder.php
```diff
@@ -122,14 +169,19 @@ public function build()
         $this->defaultOptionals();
         $this->validate();
 
-        return new UserRefreshCredentials(
-            null,
-            [
+        if ($this->jsonKeyFilePath !== null) {
```
Raibaz: Shouldn't we check for both the `jsonKeyFilePath` and the `scopes` here, since they are both required for service accounts?

Raibaz (changes requested):

src/Google/Ads/GoogleAds/Lib/OAuth2TokenBuilder.php
fiboknacky: It's already done in validate();
If that requirement fails, it will not go until this line.
What we need is just a simple check what mode a user is using.
We could define a boolean, a string that indicates that during we check in validate() too, but it's obviously overkill.

Raibaz: Got it, makes sense.

Raibaz (approved):

tests/Google/Ads/GoogleAds/Lib/OAuth2TokenBuilderTest.php
```diff
@@ -65,4 +70,117 @@ public function testBuildFailsWhenMissingRequiredValuesForInstAppOrWebFlow()
...
+                'OAUTH2',
+                $this->jsonKeyFilePath
+            ], [
+                'scopes',
+                'OAUTH2',
+                'https://www.googleapis.com/auth/adwords'
+            ], [
+                'impersonatedEmail',
+                'OAUTH2',
+                'gaapi@gmail.com'
```
sundquist: This is not an actual email, is it? I would hate for someone to get spammed if a bot or a copy-paste person reads this. I know I'm being cautious, but if possible, something completely non-existent would be nice.

tests/Google/Ads/GoogleAds/Lib/OAuth2TokenBuilderTest.php
```diff
@@ -65,4 +70,117 @@ public function testBuildFailsWhenMissingRequiredValuesForInstAppOrWebFlow()
...
+
+    /**
+     * @covers \Google\Ads\GoogleAds\Lib\OAuth2TokenBuilder::build
+     */
+    public function testBuildWithServiceAccountFlowUsingImpersonation()
+    {
+        $tokenFetcher = $this->oAuth2TokenBuilder
+            ->withJsonKeyFilePath($this->jsonKeyFilePath)
+            ->withScopes('https://www.googleapis.com/auth/adwords')
+            ->withImpersonatedEmail('gaapi@gmail.com')
```
sundquist: Same as the comment above about concerns that this mail be real.

sundquist (changes requested):
Very minor comments, but otherwise approved.

sundquist (approved):

tests/Google/Ads/GoogleAds/Lib/OAuth2TokenBuilderTest.php
fiboknacky: Good point but I really don't know too. We may try creating a new email using that one but it'd be a bit overkill in this case, so I changed to really long one that shouldn't be used by anyone instead. :)

fiboknacky:
The Kokoro presubmit fails for the style of some tests and examples only, which should be fine, because that's not the part that I modify here.
Travis checks seem to be fine.

PR #127, opened 2019-06-17
Set max response message and metadata size.

fiboknacky:
Set max response message and metadata size accordingly to prevent edge-case problems.

### DIFF
--- a/src/Google/Ads/GoogleAds/Lib/V1/ServiceClientFactoryTrait.php
+++ b/src/Google/Ads/GoogleAds/Lib/V1/ServiceClientFactoryTrait.php
@@ -149,6 +149,19 @@ trait ServiceClientFactoryTrait
         if (!empty($this->getEndpoint())) {
             $clientOptions += [self::$SERVICE_ADDRESS_KEY => $this->getEndpoint()];
         }
+
+        $clientOptions['transportConfig'] = [
+            'grpc' => [
+                'stubOpts' => [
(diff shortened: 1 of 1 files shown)


src/Google/Ads/GoogleAds/Lib/V1/ServiceClientFactoryTrait.php
```diff
@@ -164,6 +164,17 @@ public function getGoogleAdsClientOptions()
             ];
         }
 
+        $clientOptions['transportConfig']['grpc'] += [
```
michaelbausor: What will happen here if `$clientOptions['transportConfig']['grpc']` is not set when this is called? (Which presumably could happen if `$this->getLogger()` returned empty above?)

Unfortunately there isn't a nice, clean way of merging PHP arrays (that I know of) that will avoid wiping out possible existing configuration. One approach might be to set this setting *before* the conditional setting of intercepts, because this will always be set. So your code might look like:

```php
$clientOptions['transportConfig'] = [
  'grpc' => [
    'stubOpts' => [
      // ...
    ]
  ]
];
if (!empty($this->getLogger())) {
  // ...
  $clientOptions['transportConfig']['grpc'] += ['interceptors' => [$googleAdsLoggingInterceptor]];
}
```

fiboknacky: Thanks for spotting this.

PR #129, opened 2019-06-17
Log level defaults and validation

Raibaz:
This sets a default in case the log level is not set in the configuration file and adds some format validation to make sure invalid values are not being set.

Additionally, it also adds some defensive code to make sure the proper log level is being used when logging messages.

### DIFF
--- a/src/Google/Ads/GoogleAds/Lib/V1/GoogleAdsClientBuilder.php
+++ b/src/Google/Ads/GoogleAds/Lib/V1/GoogleAdsClientBuilder.php
@@ -246,7 +246,7 @@ final class GoogleAdsClientBuilder implements GoogleAdsBuilder
 
         if (is_null($this->logLevel)) {
             $this->logLevel = LogLevel::INFO;
-        } elseif (!defined('Psr\Log\LogLevel::' . $this->logLevel)) {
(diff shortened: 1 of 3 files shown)


src/Google/Ads/GoogleAds/Lib/V1/GoogleAdsClientBuilder.php
```diff
@@ -242,6 +243,12 @@ public function validate()
                 'OAuth2 authentication credentials must not be null.'
             );
         }
+
+        if (is_null($this->logLevel)) {
+            $this->logLevel = LogLevel::INFO;
+        } elseif (!defined('Psr\Log\LogLevel::' . $this->logLevel)) {
```
fiboknacky: Will this work with the case when people specify small letters as well?

Raibaz: It didn't because I wasn't sure we should allow lowercase values for the log level.

However, it was easy to allow them, so I just added support for that.

fiboknacky (approved):

Raibaz:
Added @AnashOommen for review as @PierrickVoulet is on vacation.

fiboknacky:
I'll merge this now as we want to do another release very soon.
@AnashOommen  The PR is not complex, but if you see anything you want to change, please let us know.

PR #135, opened 2019-06-24
Fixed LogMessageFormatterTest to work with the current issue of Protobuf native extension

fiboknacky:
The current version of Protobuf library converts Int64 as Number, not String.
This modification will allow both users of Protobuf native and PHP extensions able to run this PHPUnit test.


Raibaz (approved):

michaelbausor (approved):
LGTM. Suggest filing an issue on your repo that links to https://github.com/protocolbuffers/protobuf/issues/5663 so that you can revert this change once the issue is fixed in protobuf.

PR #139, opened 2019-06-27
Added the instantiate classes test

Raibaz:
This adds a very basic test to make sure that all meaningful classes included in the codebase can be instantiated.

It does so by getting all .php files under the `src` directory and trying to instantiate the corresponding PHP class.

It filters out some of the files that don't really need to be checked, like the classes containing `metadata` or `Testing` in the namespace, or the `_config.php` files, and it also does not try to instantiate non-instantiable classes like interfaces or traits.

This can be helpful in preventing classes containing syntax errors from being released.

### DIFF
--- a/tests/Google/Ads/GoogleAds/InstantiateClassesTest.php
+++ b/tests/Google/Ads/GoogleAds/InstantiateClassesTest.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Copyright 2018 Google LLC
+ * Copyright 2019 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/tests/Google/Ads/GoogleAds/InstantiateClassesTest.php
+++ b/tests/Google/Ads/GoogleAds/InstantiateClassesTest.php
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright 2018, Google Inc.
+ * Copyright 2019, Google Inc.
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
--- a/tests/Google/Ads/GoogleAds/InstantiateClassesTest.php
+++ b/tests/Google/Ads/GoogleAds/InstantiateClassesTest.php
@@ -1,33 +1,18 @@
 <?php
-/*
- * Copyright 2019, Google Inc.
- * All rights reserved.
+/**
+ * Copyright 2018 Google LLC
  *
- * Redistribution and use in source and binary forms, with or without
- * modification, are permitted provided that the following conditions are
- * met:
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
  *
- *     * Redistributions of source code must retain the above copyright
- * notice, this list of conditions and the following disclaimer.
- *     * Redistributions in binary form must reproduce the above
- * copyright notice, this list of conditions and the following disclaimer
- * in the documentation and/or other materials provided with the
- * distribution.
- *     * Neither the name of Google Inc. nor the names of its
- * contributors may be used to endorse or promote products derived from
- * this software without specific prior written permission.
+ *     https://www.apache.org/licenses/LICENSE-2.0
  *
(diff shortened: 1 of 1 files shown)


tests/Google/Ads/GoogleAds/InstantiateClassesTest.php
```diff
@@ -0,0 +1,88 @@
+<?php
+/*
+ * Copyright 2018, Google Inc.
```
fiboknacky: Need updates to use the latest copyright.

tests/Google/Ads/GoogleAds/InstantiateClassesTest.php
```diff
@@ -0,0 +1,88 @@
...
+            return;
+        } catch (ValidationException $exception) {
+            //Disregard
+            return;
+        }
+
+        $this->assertNotNull($instance);
+    }
+
+    public function classesProvider()
```
fiboknacky: How long does this test spend?

fiboknacky (commented):
What about focusing on the generated classes?
In that way, we probably can save some time looking in the metadata folder.

tests/Google/Ads/GoogleAds/InstantiateClassesTest.php
Raibaz: Done.

tests/Google/Ads/GoogleAds/InstantiateClassesTest.php
Raibaz: On my Mac, the whole test suite including this test takes 3.5 seconds to execute, while this test only takes 1.2 seconds.

Raibaz:
IMHO, since this test takes just a bit more than a second to execute, we can leave it as it is and try to instantiate all classes rather than focusing only on the generated classes.

We can always filter the non-generated classes out later in case we need to shave some milliseconds out of the test suite execution time, but i wouldn't expect a big improvement from that since the generated classes are way more than the non-generated ones.

WDYT?

fiboknacky:
> IMHO, since this test takes just a bit more than a second to execute, we can leave it as it is and try to instantiate all classes rather than focusing only on the generated classes.
> 
> We can always filter the non-generated classes out later in case we need to shave some milliseconds out of the test suite execution time, but i wouldn't expect a big improvement from that since the generated classes are way more than the non-generated ones.
> 
> WDYT?

SGTM. I just think about the metadata/ folder. If I understand correctly, you're also removing them from your test too, right? Perhaps, it's a small improvement, so I'd not be picky about that. But just thought that if we can specify the directory in a bit more specific way, like, src/Google/Ads/GoogleAds/V*, it'd be cleaner. 
(Of course we can include those in Util/ and Lib/ just in case)

But this is up to you. I agree that it wouldn't achieve much improvement in time because what's in V* are probably much more than the other locations.

tests/Google/Ads/GoogleAds/InstantiateClassesTest.php
```diff
@@ -0,0 +1,88 @@
+<?php
+/*
+ * Copyright 2019, Google Inc.
```
fiboknacky: I believe this should be LLC already.
Could you copy from our latest file like in examples/?

fiboknacky (changes requested):

tests/Google/Ads/GoogleAds/InstantiateClassesTest.php
Raibaz: Ow, you're right, I copied the whole copyright block from an example, should be ok now.

Thanks for spotting.

tests/Google/Ads/GoogleAds/InstantiateClassesTest.php
```diff
@@ -1,33 +1,18 @@
 <?php
-/*
- * Copyright 2019, Google Inc.
- * All rights reserved.
+/**
+ * Copyright 2018 Google LLC
```
fiboknacky: Nah. It needs to be 2019. :)

fiboknacky (changes requested):

tests/Google/Ads/GoogleAds/InstantiateClassesTest.php
Raibaz: Stupid copy/paste :)

fiboknacky (approved):

PR #151, opened 2019-07-08
Added HandleKeywordPolicyViolations example

fiboknacky:


examples/ErrorHandling/HandleKeywordPolicyViolations.php
```diff
@@ -0,0 +1,261 @@
...
+            $response = $adGroupCriterionServiceClient->mutateAdGroupCriteria(
+                $customerId,
+                [$adGroupCriterionOperation]
+            );
+            printf(
+                "Added a keyword with resource name '%s'.%s",
+                $response->getResults()[0]->getResourceName(),
+                PHP_EOL
+            );
+            return;
```
Raibaz: It may be useful to add a comment clarifying that this return statement is never going to be executed because the keyword creation will throw an exception, it's not really easy to follow otherwise.

Raibaz (changes requested):

examples/ErrorHandling/HandleKeywordPolicyViolations.php
fiboknacky: I added a comment to explain more. Actually this line will be executed if you run the example more than once.

Raibaz (approved):

examples/ErrorHandling/HandleKeywordPolicyViolations.php
```diff
@@ -0,0 +1,263 @@
...
+            // You may hit this line when you run this example using the default keyword for the
+            // second time. This is because the keyword will already be exempted and the mutate
+            // request can thus be successful.
+            return;
+        } catch (GoogleAdsException $googleAdsException) {
+            $exemptPolicyViolationKeys = self::fetchExemptPolicyViolationKeys($googleAdsException);
+        }
+
+        // Try sending exemption requests for creating a keyword. However, if your keyword contains
+        // many policy violations, but not all of them are exemptible, the request will not be sent.
```
PierrickVoulet: This line contains more than 100 characters

examples/ErrorHandling/HandleKeywordPolicyViolations.php
```diff
@@ -0,0 +1,263 @@
...
+                PHP_EOL
+            );
+        }
+    }
+
+    /**
+     * Runs the example.
+     *
+     * @param GoogleAdsClient $googleAdsClient the Google Ads API client
+     * @param int $customerId the client customer ID without hyphens
```
PierrickVoulet: nit: "without hyphens" can be removed as the type is int

examples/ErrorHandling/HandleKeywordPolicyViolations.php
```diff
@@ -0,0 +1,263 @@
...
+            $response = $adGroupCriterionServiceClient->mutateAdGroupCriteria(
+                $customerId,
+                [$adGroupCriterionOperation]
+            );
+            printf(
+                "Added a keyword with resource name '%s'.%s",
+                $response->getResults()[0]->getResourceName(),
+                PHP_EOL
+            );
+            // You may hit this line when you run this example using the default keyword for the
```
PierrickVoulet: This is useful comment but it seems specific to a corner case only (run it twice with the same ad group id and default keyword).

Could this line also be hit if the keyword is not the default one or at first run if the given keyword has already been exempted in the past for this customer account or ad group?

PierrickVoulet: This comment focuses on the "return" line but I think it should be put right after the line that does the "mutate" instead.

examples/ErrorHandling/HandleKeywordPolicyViolations.php
```diff
@@ -0,0 +1,263 @@
...
+            // second time. This is because the keyword will already be exempted and the mutate
+            // request can thus be successful.
+            return;
+        } catch (GoogleAdsException $googleAdsException) {
+            $exemptPolicyViolationKeys = self::fetchExemptPolicyViolationKeys($googleAdsException);
```
PierrickVoulet: Should we add a check before continuing to ensure there are only policy violation errors?

examples/ErrorHandling/HandleKeywordPolicyViolations.php
```diff
@@ -0,0 +1,263 @@
...
+                $exemptPolicyViolationKeys
+            );
+        } else {
+            print "No exemption request is sent because your keyword contained some non-exemptible"
+                . " policy violations" . PHP_EOL;
+        }
+    }
+
+    /**
+     * Collects all policy violation key that can be exempted for sending exemption request later.
```
PierrickVoulet: Typos: "key" and "request" should be in plural form "keys" and "requests"

examples/ErrorHandling/HandleKeywordPolicyViolations.php
```diff
@@ -0,0 +1,263 @@
...
+                }
+            }
+        }
+        return $exemptPolicyViolationKeys;
+    }
+
+    /**
+     * Sends exemption requests for creating a keyword.
+     *
+     * @param int $customerId
```
PierrickVoulet: The documentation of all parameters is missing.

examples/ErrorHandling/HandleKeywordPolicyViolations.php
```diff
@@ -0,0 +1,263 @@
...
+        }
+    }
+
+    /**
+     * Collects all policy violation key that can be exempted for sending exemption request later.
+     *
+     * @param GoogleAdsClient $googleAdsClient the Google Ads API client
+     * @return PolicyViolationKey[] the exemptible policy violation keys
+     */
+    private static function fetchExemptPolicyViolationKeys(GoogleAdsException $googleAdsException)
```
PierrickVoulet: Extracting this information from the exception seems to be the only way to do it but could be difficult for developers. What about having this error extraction (and may be others) directly provided by the client lib?

examples/ErrorHandling/HandleKeywordPolicyViolations.php
```diff
@@ -0,0 +1,263 @@
...
+use Google\Ads\GoogleAds\V2\Enums\KeywordMatchTypeEnum\KeywordMatchType;
+use Google\Ads\GoogleAds\V2\Errors\GoogleAdsError;
+use Google\Ads\GoogleAds\V2\Resources\AdGroupCriterion;
+use Google\Ads\GoogleAds\V2\Services\AdGroupCriterionOperation;
+use Google\Ads\GoogleAds\V2\Services\AdGroupCriterionServiceClient;
+use Google\ApiCore\ApiException;
+use Google\Protobuf\StringValue;
+
+/**
+ * This example demonstrates how to request an exemption for policy violations of a keyword.
```
PierrickVoulet: Is there any complementary documentation (URLs) we could provide regarding policy violation and exemption request?

examples/ErrorHandling/HandleKeywordPolicyViolations.php
```diff
@@ -0,0 +1,263 @@
...
+ * Note that the example uses an exemptible policy-violating keyword by default. If you use a
+ * keyword that contains non-exemptible policy violations, they will not be sent for exemption
+ * request and you will still fail to create a keyword.
```
PierrickVoulet: May be also add what is the behavior if a keyword that does not violate any policy is used.

examples/ErrorHandling/HandleKeywordPolicyViolations.php
PierrickVoulet: I would add a comment to state that reaching this catch method is expected due to the violation policy.

PierrickVoulet (changes requested):

examples/ErrorHandling/HandleKeywordPolicyViolations.php
fiboknacky: Yes, it can be.
And that's why I didn't want to add this at first because it'll be verbose.
However, Mattia asked about this so I just put the comment here.

What about just explaining in the file PHPDoc and just remove this line.
It should be obvious that in order to hit this line, the previous operation should succeed.

examples/ErrorHandling/HandleKeywordPolicyViolations.php
fiboknacky: This is intentional because I don't have anything much to say. This is the same for the Java library case.

examples/ErrorHandling/HandleKeywordPolicyViolations.php
fiboknacky: Not yet. Because this example will be used in that guide,  I need to merge this example first.

examples/ErrorHandling/HandleKeywordPolicyViolations.php
fiboknacky: Add explanation in the file-level doc now, so removing this.

examples/ErrorHandling/HandleKeywordPolicyViolations.php
fiboknacky: I don't think we need it because if it's not a policy violation error, there will be no exempt policy key populated.
And then, there will be no attempt to send a request for exemption.
The example will just end with a message "No exemption request is sent because ... "

fiboknacky: I think it'd not be necessary because
1) We already explain this at the file-level doc.
2) It's not expected always partly because of the cases that you point out (normal keyword).

I'd keep it concise here. The private method PHPDoc will do its task to explain things in there too.

examples/ErrorHandling/HandleKeywordPolicyViolations.php
fiboknacky: I double-checked this and it's within 100 chars.

examples/ErrorHandling/HandleKeywordPolicyViolations.php
fiboknacky: Done.

examples/ErrorHandling/HandleKeywordPolicyViolations.php
fiboknacky: That could be P2 feature request. Thanks for suggestion.
Let's keep it as is for now.

examples/ErrorHandling/HandleKeywordPolicyViolations.php
fiboknacky: Done.

PierrickVoulet (approved):

PR #153, opened 2019-07-10
Added HandleExpandedTextAdPolicyViolations example

fiboknacky:


examples/ErrorHandling/HandleExpandedTextAdPolicyViolations.php
```diff
@@ -0,0 +1,244 @@
...
+            );
+        }
+    }
+
+    /**
+     * Runs the example.
+     *
+     * @param GoogleAdsClient $googleAdsClient the Google Ads API client
+     * @param int $customerId the client customer ID
+     * @param int $adGroupId the ad group ID to add a keyword to
```
PierrickVoulet: No keyword is added to the ad group in this example, I think you meant to write 'an expanded text ad to' instead

examples/ErrorHandling/HandleExpandedTextAdPolicyViolations.php
```diff
@@ -0,0 +1,244 @@
...
+     */
+    public static function runExample(
+        GoogleAdsClient $googleAdsClient,
+        int $customerId,
+        int $adGroupId
+    ) {
+        $adGroupResourceName =
+            new StringValue(['value' => ResourceNames::forAdGroup($customerId, $adGroupId)]);
+
+        $expandedTextAdInfo = new ExpandedTextAdInfo([
```
PierrickVoulet: nit: add a comment saying the expanded ad is created here

PierrickVoulet (changes requested):

examples/ErrorHandling/HandleExpandedTextAdPolicyViolations.php
fiboknacky: Done.

examples/ErrorHandling/HandleExpandedTextAdPolicyViolations.php
```diff
@@ -0,0 +1,245 @@
...
+                PHP_EOL
+            );
+            if (!is_null($error->getDetails())
+                && !is_null($error->getDetails()->getPolicyFindingDetails())) {
+                $policyFindingDetails = $error->getDetails()->getPolicyFindingDetails();
+                printf("\tPolicy finding details:%s", PHP_EOL);
+
+                foreach ($policyFindingDetails->getPolicyTopicEntries() as $policyTopicEntry) {
+                    /** @var PolicyTopicEntry $policyTopicEntry */
+                    $ignorablePolicyTopics[] = $policyTopicEntry->getTopic();
```
Raibaz: Is there a specific reason for returning an array of StringValues here rather than calling `getTopicUnwrapped()` and returning an array of strings?

Raibaz (changes requested):

PierrickVoulet (approved):

examples/ErrorHandling/HandleExpandedTextAdPolicyViolations.php
fiboknacky: Because the array of SringValue will be used to request exemption.
It makes no sense to get each as primitive string, just to create a StringValue to wrap it later.

Raibaz (approved):

PR #155, opened 2019-07-11
Added GoogleAdsFailure initialization and removed unused imports

Raibaz:
This fixes #154.

PR #156, opened 2019-07-12
Use more reliable way to test GoogleAdsFailure

### LINKED ISSUE #154: Broken HandlePartialFailure code example
I tried to run the code example with both v1 and v2 but reach the following (redacted) error:

```
[2019-07-10 13:36:26] google-ads.INFO: Request made: Host: "googleads.googleapis.com:443", Method: "/google.ads.googleads.v2.services.AdGroupService/MutateAdGroups", ClientCustomerId: XXX, RequestId: "YYY", IsFault: 0, FaultMessage: "None"  
Partial failures occurred. Details will be shown below.
Operation 0 succeeded: ad group with resource name 'customers/XXX/adGroups/ZZZ'.
PHP Fatal error:  Uncaught Exception: Class google.ads.googleads.v2.errors.GoogleAdsFailure hasn't been added to descriptor pool in /usr/.../google-ads-php/vendor/google/protobuf/src/Google/Protobuf/Any.php:275
Stack trace:
#0 /usr/.../google-ads-php/src/Google/Ads/GoogleAds/Util/V2/GoogleAdsFailures.php(34): Google\Protobuf\Any->unpack()
#1 /usr/.../google-ads-php/src/Google/Ads/GoogleAds/Util/V2/GoogleAdsErrors.php(87): Google\Ads\GoogleAds\Util\V2\GoogleAdsFailures::fromAny(Object(Google\Protobuf\Any))
#2 /usr/.../google-ads-php/examples/ErrorHandling/HandlePartialFailure.php(211): Google\Ads\GoogleAds\Util\V2\GoogleAdsErrors::fromStatus(1, Object(Google\Rpc\Status))
#3 /usr/.../google-ads-php/examples/ErrorHandling/HandlePartialFailure.php(117): Google\Ads\GoogleAds\Examples\ErrorHandling\HandlePartialFailure::printResults(Object(Google in /usr/.../google-ads-php/vendor/google/protobuf/src/Google/Protobuf/Any.php on line 275
```

PierrickVoulet:
This is to enhance #155 which fixes #154.

### DIFF
--- a/src/Google/Ads/GoogleAds/Util/V1/GoogleAdsFailures.php
+++ b/src/Google/Ads/GoogleAds/Util/V1/GoogleAdsFailures.php
@@ -18,14 +18,12 @@
 namespace Google\Ads\GoogleAds\Util\V1;
 
 use Google\Protobuf\Any;
+use Google\Protobuf\DescriptorPool;
 use Google\Rpc\Status;
 use Google\Ads\GoogleAds\V1\Errors\GoogleAdsFailure;
 
 final class GoogleAdsFailures
 {
-    
-    private static $unusedGoogleAdsFailure;
-
     /**
      * Unpacks a single GoogleAdsFailure from an Any instance.
      *
@@ -36,8 +34,10 @@ final class GoogleAdsFailures
     {
         // This initialization is needed to populate the descriptor pool with the GoogleAdsFailure
         // class and prevent exceptions from being thrown.
-        if (!isset(self::$unusedGoogleAdsFailure)) {
-            self::$unusedGoogleAdsFailure = new GoogleAdsFailure();
+        if (is_null(
+            DescriptorPool::getGeneratedPool()->getDescriptorByClassName(GoogleAdsFailure::class)
+        )) {
+            new GoogleAdsFailure();
         }
 
         $ret = $any->unpack();
--- a/src/Google/Ads/GoogleAds/Util/V2/GoogleAdsFailures.php
+++ b/src/Google/Ads/GoogleAds/Util/V2/GoogleAdsFailures.php
@@ -18,14 +18,12 @@
 namespace Google\Ads\GoogleAds\Util\V2;
 
 use Google\Protobuf\Any;
+use Google\Protobuf\DescriptorPool;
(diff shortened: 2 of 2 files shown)


src/Google/Ads/GoogleAds/Util/V1/GoogleAdsFailures.php
```diff
@@ -36,8 +34,10 @@ public static function fromAny(Any $any)
     {
         // This initialization is needed to populate the descriptor pool with the GoogleAdsFailure
         // class and prevent exceptions from being thrown.
-        if (!isset(self::$unusedGoogleAdsFailure)) {
-            self::$unusedGoogleAdsFailure = new GoogleAdsFailure();
+        if (is_null(
+            DescriptorPool::getGeneratedPool()->getDescriptorByClassName(GoogleAdsFailure::class)
```
fiboknacky: 1) Have you confirmed with the Protobuf team that this is a reliable way to fix?
2) The generated pool will always retain all descriptors in a run time until the program finishes?

PierrickVoulet: 1. I have not confirmed directly with Protobuf team, this is based on a discussion I got with Bob.
2. I do not have the answer to this question, actually the purpose of this edit is to make sure we do not have to know: I basically check and instantiate a new object anytime I cannot find the descriptor that is necessary in the pool that is being used (singleton) by Any to unpack. This is board-effect free as it works regardless of how PHP, Protobuf or any other dependencies behave at runtime. The current test assumes that instantiating once the object will be enough to load the descriptor pool forever, this may be true for most cases today but there might be corner cases we do not know and/or it could change in future for unforeseeable reasons.

Raibaz: Point no. 2 here seems pretty solid to me, it really looks better to set the descriptor in the pool directly rather than relying on variable instantiation doing so implicitly.

fiboknacky: SGTM.

PR #157, opened 2019-07-12
Add protobuf impl arg in dockerfile and comments

PierrickVoulet:


Dockerfile
```diff
@@ -1,15 +1,23 @@
 FROM php:7.1-apache
 
+ARG USE_C_PROTOBUF=false
```
Raibaz: Why is this false by default? I thought we'd switch to using the C extension by default.

PierrickVoulet: Good point. I should have detailed that the purpose of this PR is to add the build argument only: I set the default as false to not introduce any breaking changes.

The change to Protobuf C will be done by other PR(s) and changing the value to true will be part of it.

Raibaz: Got it, thanks.

Dockerfile
```diff
@@ -1,15 +1,23 @@
...
+
 RUN apt-get update && apt-get install -y libxml2 zlib1g-dev git unzip
 
+# Install PHP extension(s) required for development
```
fiboknacky: Nit: Please end all your code comments with a period.

Dockerfile
```diff
@@ -1,15 +1,23 @@
...
+# Install and configure Composer
 RUN curl -sS https://getcomposer.org/installer | php
 RUN mv composer.phar /usr/local/bin/composer
 
+# Install and configure the gRPC extension
 RUN pecl install grpc
 RUN echo 'extension=grpc.so' >> $PHP_INI_DIR/conf.d/grpc.ini
 
+# Install and configure the C implementation of Protobuf extension if needed
+RUN if [ "$USE_C_PROTOBUF" = "false" ]; then echo 'Use PHP implementation of Protobuf'; else echo 'Use C implementation of Protobuf'; pecl install protobuf; echo 'extension=protobuf.so' >> $PHP_INI_DIR/conf.d/protobuf.ini; fi
```
fiboknacky: Nit: "Using" instead of "Use" would be a bit easy to understand when logged as a text.

Dockerfile
PierrickVoulet: Done

Dockerfile
PierrickVoulet: Done

PR #158, opened 2019-07-12
Add documentation related to Protobuf

PierrickVoulet:


README.md
```diff
@@ -182,6 +197,36 @@ $googleAdsClient = (new GoogleAdsClientBuilder())
...
+[Protobuf](https://developers.google.com/protocol-buffers/docs/overview) is used by
+[gRPC](https://grpc.io/about/), a core dependency of the Google Ads API for request and
+response transport.
+
+There are two implementations available for PHP
+
+1. C (native): for better performance
+1. PHP (pure): to get started easily
+
+By default implementation of Protobuf used is C. This might not be suitable for some users
```
Raibaz: Nit: how about "By default, the C implementation is used"?

README.md
```diff
@@ -182,6 +197,36 @@ $googleAdsClient = (new GoogleAdsClientBuilder())
...
+who then need to use the PHP implementation instead. The PHP implementation
```
Raibaz: Should we provide some examples of use cases where the PHP implementation is needed?

PierrickVoulet: The main reason is security but there might be others. @fiboknacky do you have any good resources to share regarding this?

README.md
PierrickVoulet: Done

README.md
```diff
@@ -182,6 +197,36 @@ $googleAdsClient = (new GoogleAdsClientBuilder())
...
+[Protobuf](https://developers.google.com/protocol-buffers/docs/overview) is used by
+[gRPC](https://grpc.io/about/), a core dependency of the Google Ads API for request and
+response transport.
+
+There are two implementations available for PHP
+
+1. C (native): for better performance
+1. PHP (pure): to get started easily
+
+By default, the C implementation is used. This might not be suitable for some users
```
fiboknacky: I doubt if "By default" here is correct in the meaning.
Would be a bit more accurate to say that "if you follow the steps above for installing C implementation...", for example.

The reason is we always have beginners who somehow skip doing some steps and in that case, they would end up using PHP implementation anyway.
So for them, it's probably not quite "by default".

README.md
fiboknacky: I cannot think of other good usages too.
The main reason would be simplicity and in some web servers, it's probably hard to install C implementation, which requires you to install via sudo

README.md
```diff
@@ -182,6 +197,36 @@ $googleAdsClient = (new GoogleAdsClientBuilder())
...
+who then need to use the PHP implementation instead. The PHP implementation
+is automatically installed like any other [`googleapis/gax-php`](https://github.com/googleapis/gax-php)
+dependency via Composer but it is only used if the C implementation is not
+installed and configured.
+
+### Determine which  implementation is being used
```
fiboknacky: Extra space between "which" and "implementation"

README.md
PierrickVoulet: Ok, I am using the security example here then, it does not have to be exhaustive anyway.

README.md
PierrickVoulet: DONE

README.md
PierrickVoulet: Good point, done.

PierrickVoulet:
> Please move most changes to DevSite (https://developers.google.com/google-ads/api/docs/client-libs/php/).
> Mentions things only briefly here.

I am working on the DevSite edition and will edit this PR accordingly when it is live.

fiboknacky:
OK. Ping me again once your DevSite change goes online.

PierrickVoulet:
Edit to come: only keep instructions in the getting started section to install the C implementation, other will be replaced by a reference to the DevSite guide (WIP)

README.md
```diff
@@ -12,6 +12,22 @@ API](https://developers.google.com/google-ads/api/docs/start).
 
 ## Requirements
 
+*   Two PHP extensions need to be installed to use the library at runtime
+    *   gRPC: follow the section `Install the gRPC PHP extension`
+    of [this page](https://grpc.io/docs/quickstart/php.html)
```
fiboknacky: Hanging indent should be done here too. In other words, please align "of" with "gRPC" in the previous line.

README.md
```diff
@@ -12,6 +12,22 @@ API](https://developers.google.com/google-ads/api/docs/start).
 
 ## Requirements
 
+*   Two PHP extensions need to be installed to use the library at runtime
```
fiboknacky: I don't think we need to put this as a first line. It can be grouped under the bullet you've just removed "System requirements and dependencies..."

PierrickVoulet: Done

README.md
PierrickVoulet: Done

README.md
```diff
@@ -12,6 +12,22 @@ API](https://developers.google.com/google-ads/api/docs/start).
...
+    *   gRPC: follow the section `Install the gRPC PHP extension`
+    of [this page](https://grpc.io/docs/quickstart/php.html)
+        1.  Install the extension using the command `sudo pecl install grpc`.
+        1.  Add a line `extension=grpc.so` to the `php.ini` file.
+        1.  Run `php -i | grep grpc` in a terminal: it is well installed
+        and configured if it returns something
+    *   Protobuf: follow the section `C implementation` under
```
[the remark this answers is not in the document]
PierrickVoulet: Done

README.md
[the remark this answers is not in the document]
fiboknacky: I see. Thanks for doing research on this.
The suggestion in composer.json looks good to me.
Could you either use do it here or another PR (in case you want to submit this PR quickly)?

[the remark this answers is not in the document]
PierrickVoulet: No worries, I will be sending an update in this PR as I also need to change the default value of the "user C protobuf" argument in the Dockerfile anyway.

PierrickVoulet:
@fiboknacky I requested a review again from you because I changed the wording in the suggestions.

PR #160, opened 2019-07-16
Added AddSmartDisplayAd example.

fiboknacky:


examples/AdvancedOperations/AddSmartDisplayAd.php
```diff
@@ -0,0 +1,434 @@
...
+     *     square marketing image asset
+     */
+    private static function createResponsiveDisplayAd(
+        GoogleAdsClient $googleAdsClient,
+        int $customerId,
+        string $adGroupResourceName,
+        string $marketingImageAssetResourceName = null,
+        string $squareMarketingImageAssetResourceName = null
+    ) {
+        // Creates a new image asset for marketing image and square marketing image if there is no
```
Raibaz: Nit: "if there are no assets' resource names"

examples/AdvancedOperations/AddSmartDisplayAd.php
```diff
@@ -0,0 +1,434 @@
...
+                $customerId,
+                self::SQUARE_MARKETING_IMAGE_URL,
+                'Square Marketing Image'
+            );
+
+        // Creates an ad group ad with the ad info related to the responsive display ad.
+        $adGroupAd = new AdGroupAd([
+            'ad_group' => $adGroupResourceName,
+            'status' => AdGroupAdStatus::PAUSED,
+            'ad' => new Ad([
```
Raibaz: The ad object here is pretty big, so the AdGroupAd creation doesn't fit in a single screen.

How about extracting it to a separate variable?

Raibaz (changes requested):

examples/AdvancedOperations/AddSmartDisplayAd.php
fiboknacky: Done.

Raibaz (approved):

examples/AdvancedOperations/AddSmartDisplayAd.php
```diff
@@ -0,0 +1,436 @@
...
+
+        // Issues a mutate request to add campaign budgets.
+        $campaignBudgetServiceClient = $googleAdsClient->getCampaignBudgetServiceClient();
+        /** @var MutateCampaignBudgetsResponse $campaignBudgetResponse */
+        $campaignBudgetResponse = $campaignBudgetServiceClient->mutateCampaignBudgets(
+            $customerId,
+            [$campaignBudgetOperation]
+        );
+
+        $campaignBudgetResourceName = $campaignBudgetResponse->getResults()[0]->getResourceName();
```
PierrickVoulet: nit: add comments about printing out some information about added elements

examples/AdvancedOperations/AddSmartDisplayAd.php
```diff
@@ -0,0 +1,436 @@
...
+            'long_headline' => new AdTextAsset([
+                'text' => new StringValue(['value' => 'Travel the World'])
+            ]),
+            'descriptions' => [
+                new AdTextAsset([
+                    'text' => new StringValue(['value' => 'Take to the air!'])
+                ])
+            ],
+            'business_name' => new StringValue(['value' => 'Google']),
+            // Sets the marketing image and square marketing image to the previously created
```
PierrickVoulet: These might not have been created previously, I would remove 'to the previously created image assets' part

examples/AdvancedOperations/AddSmartDisplayAd.php
```diff
@@ -0,0 +1,436 @@
...
+    private static function createResponsiveDisplayAd(
+        GoogleAdsClient $googleAdsClient,
+        int $customerId,
+        string $adGroupResourceName,
+        string $marketingImageAssetResourceName = null,
+        string $squareMarketingImageAssetResourceName = null
+    ) {
+        // Creates a new image asset for marketing image and square marketing image if there are no
+        // assets' resource names specified.
+        $marketingImageAssetResourceName = $marketingImageAssetResourceName ?:
```
PierrickVoulet: Why not creating 'AdImageAsset' instances directly from here? This would reduce the size of the responsiveDisplayAdInfo variable construction statement.

PierrickVoulet (changes requested):

examples/AdvancedOperations/AddSmartDisplayAd.php
```diff
@@ -0,0 +1,436 @@
...
+use Google\Protobuf\BytesValue;
+use Google\Protobuf\Int64Value;
+use Google\Protobuf\StringValue;
+
+/**
+ * This example adds a Smart Display campaign, an ad group and a responsive display ad.
+ * More information about Smart Display campaigns can be found at
+ * https://support.google.com/google-ads/answer/7020281.
+ *
+ * IMPORTANT: The AssetService requires you to reuse what you've uploaded previously. Therefore,
```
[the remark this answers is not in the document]
fiboknacky: > This might become a blocking issue when there will be more than one code example using assets that loads the same images

Why? We can share assets and images across ads and across code examples. Could you elaborate?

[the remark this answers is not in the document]
fiboknacky: Note that 2) is not feasible especially for images, AssetService translates the URL we've uploaded into something else (I guess it normalizes all the URLs), so you cannot just use the URL you use to query.

examples/AdvancedOperations/AddSmartDisplayAd.php
fiboknacky: No, I think this is correct.
It's either previously (long time ago) created or has just been previously created in the above lines by calling createImageAsset() function. :)
So, this is literally correct.

examples/AdvancedOperations/AddSmartDisplayAd.php
fiboknacky: I believe you cannot do that. For images, you would need to use AssetService to create image assets first unlike AdTextAsset.

fiboknacky: Sorry I misunderstand your comment, but I think it's good as is. We need to create AdTextAsset, so this is more consistent and easier to read.

But it seems AdImageAsset instantiation can be shortened into one line. I'll do that soon.

examples/AdvancedOperations/AddSmartDisplayAd.php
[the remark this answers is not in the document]
PierrickVoulet: > > This might become a blocking issue when there will be more than one code example using assets that loads the same images
> 
> Why? We can share assets and images across ads and across code examples. Could you elaborate?

If I understand well your comment, it would be impossible to create multiple assets using the same URL (image bytes). So if I run the AddSmartDisplayAd and then the AddGmailAd code examples both with default image URLs, the AddGmailAd code example will fail because the same image URLs are being used.

I am suggesting to somehow make it smoother for the developers either by adding comments or to workaround this constraint programmatically in all the code examples that share the same image URLs.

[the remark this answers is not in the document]
fiboknacky: > If I understand well your comment, it would be impossible to create multiple assets using the same URL (image bytes). So if I run the AddSmartDisplayAd and then the AddGmailAd code examples both with default image URLs, the AddGmailAd code example will fail because the same image URLs have been used before.

I believe that's a different system.
If AddGmailAd uses assets, the user should specify the existing asset resource names.
If not (like using a MediaService's created image), they're not related and this shouldn't block them from uploading the same image.

examples/AdvancedOperations/AddSmartDisplayAd.php
```diff
@@ -0,0 +1,436 @@
...
+            [$adGroupAdOperation]
+        );
+
+        // Prints out some information about the newly created ad.
+        $adGroupAdResourceName = $adGroupAdResponse->getResults()[0]->getResourceName();
+        printf("Added ad group ad named '%s'.%s", $adGroupAdResourceName, PHP_EOL);
+    }
+
+    /**
+     * Creates an image asset to be used for creating ads.
```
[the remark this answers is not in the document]
fiboknacky: Yes. Assets are for creating ads.

fiboknacky:
@PierrickVoulet  Done all changes. Please check

examples/AdvancedOperations/AddSmartDisplayAd.php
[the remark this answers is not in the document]
PierrickVoulet: Ok thanks for clarifying.

PierrickVoulet (approved):

PR #162, opened 2019-07-18
Change the default protobuf implementation used in the Docker image

PierrickVoulet:


PierrickVoulet:
This is breaking change in the context of the Docker image so I am not sure if it makes sense to make a release for this alone?

Issue #138, opened 2019-06-25
Make the test modified in #135 more specific again


fiboknacky:
#135 was done to cope with [the bug of the Protobuf extension](https://github.com/protocolbuffers/protobuf/issues/5663), still pending fix.
The PR made all users of PHP and C extensions be able to run their tests.
This resulted in a bit more relaxed test.

Once the bug is fixed, we should revert back to use more specific test again.

fiboknacky:
Fixed by #170

PR #170, opened 2019-07-29
Revert the fix for LogMessageFormatterTest.

fiboknacky:
This is a revert of the temporary fix for the protobuf library's [issue](https://github.com/protocolbuffers/protobuf/issues/5663).
Now, the original issue is fixed, so we can revert the test to be more specific again.

This is a fix for #138.


fiboknacky:
Note that somehow the v2 test is the old version (the specific one) already, so we don't need to do the revert for it.

PR #171, opened 2019-07-29
Fixed @cover of some tests + added support for code coverage.

fiboknacky:
<filter><whitelist> is needed for code coverage module to work properly.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -8,4 +8,14 @@
     <logging>
         <log type="coverage-text" target="php://stdout"/>
     </logging>
+
+    <filter>
+        <whitelist>
+            <directory>src</directory>
+            <exclude>
+                <directory>src/Google/Ads/GoogleAds/V*</directory>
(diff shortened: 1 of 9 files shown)


phpunit.xml.dist
```diff
@@ -8,4 +8,14 @@
...
     </logging>
-</phpunit>
\ No newline at end of file
+
+    <filter>
+        <whitelist>
+            <directory>src</directory>
+            <exclude>
+                <directory>src/Google/Ads/GoogleAds/V*</directory>
+                <directory>src/Google/Ads/GoogleAds/Util/Testing</directory>
```
PierrickVoulet: Should we also exclude "src/Google/Ads/GoogleAds/Lib/Testing"?

PierrickVoulet (approved):

phpunit.xml.dist
fiboknacky: Done.

PR #175, opened 2019-07-30
Added support for Travis and update README.md to include stats

fiboknacky:
.yml config file was tested and ready to be reviewed.


.travis.yml
```diff
@@ -0,0 +1,24 @@
...
+php:
+  - 7.1
+  - 7.2
+  - 7.3
+
+branches:
+  only:
+  - /.*/
+
+env:
```
PierrickVoulet: A lot of logic of this file duplicates what we already maintain in our docker file. Have you considered using the [docker](https://docs.travis-ci.com/user/docker/#building-a-docker-image-from-a-dockerfile) feature of Travis?

My feeling is that this is the perfect occasion to upgrade our docker file for all CI purposes.

fiboknacky: I'm thinking about that too. But I think the purpose of Travis is pretty overlapping to Docker itself, that is to create a kind of virtual machine and run the CI jobs in there.
Using docker inside Travis is probably fine too, but it would suit more complicated situation where Travis doesn't support.

Besides that, we need to support multiple PHP versions, which surely can be done in Docker.
But that will make Dockerfile more complicated, which makes beginners harder to grasp.

The main purpose of Dockerfile in here would be to get people easily started, and that's why we should keep the Dockerfile as simple as possible.
For now, let's Travis handle multiple PHP versions for testing purposes.

fiboknacky: A bit minor advantage is that when you use native Travis, you can see [the reports created for each PHP version separated nicely in the UI](https://travis-ci.org/googleads/google-ads-php/builds/566042402?utm_source=github_status&utm_medium=notification)
I'm not saying that it's impossible for Docker. Probably we can configure it to show something similar as well, but we can get this for free by Travis infra.

If you're fine, I'd propose to implement Travis like this first.
In the future, if we find we need more complicated settings in the VM, we can always get back to Docker.

PierrickVoulet: SGTM

PierrickVoulet (approved):

PR #177, opened 2019-07-31
Removed options from filter_var() as it's deprecated in PHP 7.3

fiboknacky:
Removal of the options doesn't affect any earlier PHP versions too. It's just redundant and PHP 7.3 decides to throw an error instead.


PierrickVoulet (approved):
This [documentation](https://wiki.php.net/rfc/deprecations_php_7_3#filter_flag_scheme_required_and_filter_flag_host_required) confirms that these are deprecated because useless, I quote "As of PHP 5.2.1 FILTER_VALIDATE_URL implies FILTER_FLAG_SCHEME_REQUIRED and FILTER_FLAG_HOST_REQUIRED"

PR #178, opened 2019-07-31
Updated type hint of examples' function variables

fiboknacky:
- "client customer ID" is replaced with just "customer ID", which is the often used word in Google Ads API. "without hyphens" is also stripped because I made all of them of type hint "int" 
- Also fixed some minor issues in the example files.
- Recommendation ID is technically a string by design, so I leave it as is.

### DIFF
--- a/examples/Recommendations/ApplyRecommendation.php
+++ b/examples/Recommendations/ApplyRecommendation.php
@@ -98,13 +98,13 @@ class ApplyRecommendation
      * Runs the example.
      *
      * @param GoogleAdsClient $googleAdsClient the Google Ads API client
-     * @param int $customerId the client customer ID without hyphens
+     * @param int $customerId the customer ID
      * @param string $recommendationId the recommendation ID to apply
      */
     public static function runExample(
         GoogleAdsClient $googleAdsClient,
-        $customerId,
-        $recommendationId
+        int $customerId,
+        string $recommendationId
     ) {
         $recommendationResourceName =
             ResourceNames::forRecommendation($customerId, $recommendationId);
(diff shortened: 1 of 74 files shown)


examples/Recommendations/ApplyRecommendation.php
```diff
@@ -98,13 +98,13 @@ public static function main()
...
      */
     public static function runExample(
         GoogleAdsClient $googleAdsClient,
-        $customerId,
-        $recommendationId
+        int $customerId,
+        string $recommendationId
     ) {
         $recommendationResourceName =
             ResourceNames::forRecommendation($customerId, $recommendationId);
```
PierrickVoulet: `ResourceNames::forRecommendation` takes an `int` typed variable as second parameter, not a `string`

examples/Recommendations/DismissRecommendation.php
```diff
@@ -98,13 +98,13 @@ public static function main()
...
      */
     public static function runExample(
         GoogleAdsClient $googleAdsClient,
-        $customerId,
-        $recommendationId
+        int $customerId,
+        string $recommendationId
     ) {
         $recommendationResourceName =
             ResourceNames::forRecommendation($customerId, $recommendationId);
```
PierrickVoulet: `ResourceNames::forRecommendation` takes an `int` typed variable as second parameter, not a `string`

PierrickVoulet (changes requested):

examples/Recommendations/ApplyRecommendation.php
fiboknacky: forRecommendation() is wrong. I just fixed that in this PR as well. 
Thanks for catching.

PierrickVoulet (approved):

Raibaz (approved):

PR #179, opened 2019-08-01
Added campaign experiments examples.

fiboknacky:
Change-Id: I84c54afb519ce42d0e84b7b058e6cecb79ed3400

### DIFF
--- a/examples/CampaignManagement/CreateCampaignExperiment.php
+++ b/examples/CampaignManagement/CreateCampaignExperiment.php
@@ -1,0 +1,163 @@
+<?php
+/**
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Ads\GoogleAds\Examples\CampaignManagement;
+
+require __DIR__ . '/../../vendor/autoload.php';
+
+use GetOpt\GetOpt;
+use Google\Ads\GoogleAds\Examples\Utils\ArgumentNames;
+use Google\Ads\GoogleAds\Examples\Utils\ArgumentParser;
+use Google\Ads\GoogleAds\Lib\OAuth2TokenBuilder;
+use Google\Ads\GoogleAds\Lib\V2\GoogleAdsClient;
+use Google\Ads\GoogleAds\Lib\V2\GoogleAdsClientBuilder;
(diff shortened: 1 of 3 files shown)


examples/CampaignManagement/CreateCampaignExperiment.php
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+
+    /**
+     * Runs the example.
+     *
+     * @param GoogleAdsClient $googleAdsClient the Google Ads API client
+     * @param int $customerId the client customer ID
+     * @param int $baseCampaignId the base campaign ID
+     * @param int $draftId the draft resource name used to create an experiment
```
PierrickVoulet: the type is int but the description refers to a resource name which is a string.

PierrickVoulet (changes requested):

examples/CampaignManagement/CreateCampaignExperiment.php
```diff
@@ -0,0 +1,163 @@
...
+use Google\Ads\GoogleAds\Util\V2\ResourceNames;
+use Google\Ads\GoogleAds\V2\Enums\CampaignExperimentTrafficSplitTypeEnum\CampaignExperimentTrafficSplitType;
+use Google\Ads\GoogleAds\V2\Errors\GoogleAdsError;
+use Google\Ads\GoogleAds\V2\Resources\CampaignExperiment;
+use Google\Ads\GoogleAds\V2\Services\GoogleAdsRow;
+use Google\ApiCore\ApiException;
+use Google\Protobuf\StringValue;
+
+/**
+ * This example illustrates how to begin creation of a campaign experiment from a draft and wait
```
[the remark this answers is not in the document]
fiboknacky: This is from the golden CL.
I guess this wording is because of the fact that the process is async, unlike other examples where you send a request to the API server, and then an object is created almost exactly the same time the result is returned to you.

For this one, you just send a request to make it begin creation process.
Then, after that, you need to poll it to see if it's finished.

examples/CampaignManagement/CreateCampaignExperiment.php
fiboknacky: This is a mistake. Done.

examples/CampaignManagement/CreateCampaignExperiment.php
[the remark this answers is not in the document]
PierrickVoulet: Makes sense

PierrickVoulet (approved):

examples/CampaignManagement/CreateCampaignExperiment.php
```diff
@@ -0,0 +1,163 @@
...
+        printf(
+            'Asynchronous request to create campaign experiment with resource name "%1$s"'
+            . ' started.%2$sWaiting until operation completes.%2$s',
+            $campaignExperimentResourceName,
+            PHP_EOL
+        );
+
+        // pollUntilComplete() implements a default back-off policy for retrying. You can tweak the
+        // retrying parameters like the maximum polling interval to use by passing them as an array
+        // to the pollUntilComplete() function. Visits the OperationResponse.php file for more
```
Raibaz: Nit: "Visit"

Raibaz (changes requested):

Raibaz (approved):

PR #181, opened 2019-08-05
Added AddGoogleMyBusinessLocationExtensions example

fiboknacky:


examples/Extensions/AddGoogleMyBusinessLocationExtensions.php
```diff
@@ -0,0 +1,252 @@
...
+            [$feedOperation]
+        );
+        $gmbFeedResourceName = $response->getResults()[0]->getResourceName();
+        printf("GMB feed created with resource name: '%s'.%s", $gmbFeedResourceName, PHP_EOL);
+        // Creates a customer feed that associates the feed with this customer for the LOCATION
+        // placeholder type.
+        $customerFeed = new CustomerFeed([
+            'feed' => new StringValue(['value' => $gmbFeedResourceName]),
+            'placeholder_types' => [PlaceholderType::LOCATION],
+            // Creates a matching function that will always evaluate to true.
```
Raibaz: Should we detail a bit further the reason why this is needed? It seems a bit unclear to me.

examples/Extensions/AddGoogleMyBusinessLocationExtensions.php
```diff
@@ -0,0 +1,252 @@
...
+        } while ($numberOfAttempts < self::MAX_CUSTOMER_FEED_ADD_ATTEMPTS
+            && is_null($addedCustomerFeed));
+
+        if (is_null($addedCustomerFeed)) {
+            throw new \RuntimeException(
+                'Could not create the customer feed after ' . self::MAX_CUSTOMER_FEED_ADD_ATTEMPTS
+                . ' attempts. Please retry the customer feed ADD operation later.'
+            );
+        }
+        // OPTIONAL: Create a campaign feed to specify which feed items to use at the campaign
```
Raibaz: Will these be implemented or are we leaving these as they are now?

Raibaz (changes requested):

examples/Extensions/AddGoogleMyBusinessLocationExtensions.php
fiboknacky: Leave as is. This is aligned with the golden CL, and it's already a long code example.

examples/Extensions/AddGoogleMyBusinessLocationExtensions.php
fiboknacky: We've never explained this more in our AdWords API guides as well.
This is migrated as is from AdWords API to the Google Ads API.
I guess it's just the internal mechanism of feed serving, that you always need a matching function.

In this case, always true === no filer at all.

Raibaz: Got it.

examples/Extensions/AddGoogleMyBusinessLocationExtensions.php
Raibaz: Ok.

Raibaz (approved):

examples/Extensions/AddGoogleMyBusinessLocationExtensions.php
```diff
@@ -0,0 +1,252 @@
...
+     * Runs the example.
+     *
+     * @param GoogleAdsClient $googleAdsClient the Google Ads API client
+     * @param int $customerId the client customer ID
+     * @param string $gmbEmailAddress the email address associated with the GMB account
+     * @param string $gmbAccessToken the access token created using the 'AdWords' scope and the
+     *     client ID and client secret of with the Cloud project associated with the GMB account
+     * @param int $businessAccountIdentifier the account number of the GMB account
+     */
+    public static function runExample(
```
PierrickVoulet: Breakdown into multiple functions (2 seems right)

examples/Utils/ArgumentNames.php
```diff
@@ -71,6 +75,8 @@ final class ArgumentNames
         self::CPC_BID_MICRO_AMOUNT => 'The CPC bid micro amount',
         self::CRITERION_ID => 'The criterion ID',
         self::CUSTOMER_ID => 'The customer ID without dashes',
+        self::GMB_ACCESS_TOKEN => 'The access token used for uploading GMB location feed data',
```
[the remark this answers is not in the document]
fiboknacky: Let's do the rename when the time comes, but in the whole AdWords API, we never need to use this except for GMB before. I don't think we're the fan of passing access token like this too, so in most places, it should be probably handled by using normal OAuth2 process.

PierrickVoulet (approved):

PR #187, opened 2019-08-14
Added support for mutate_operations in errors

Raibaz:
See the [equivalent Java PR](https://github.com/googleads/google-ads-java/pull/126) for reference.

### DIFF
--- a/tests/Google/Ads/GoogleAds/Util/V1/GoogleAdsErrorsTest.php
+++ b/tests/Google/Ads/GoogleAds/Util/V1/GoogleAdsErrorsTest.php
@@ -18,7 +18,6 @@
 namespace Google\Ads\GoogleAds\Util\V1;
 
 use Google\Protobuf\Any;
-use Google\Protobuf\Int64Value;
 use Google\Rpc\Status;
 use Google\Ads\GoogleAds\V1\Errors\ErrorLocation;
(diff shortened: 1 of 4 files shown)


fiboknacky (approved):

tests/Google/Ads/GoogleAds/Util/V1/GoogleAdsErrorsTest.php
```diff
@@ -105,4 +103,28 @@ public function testFromFailureWithInvalidOperationIndex()
         $errors = GoogleAdsErrors::fromFailure(1, $this->failure);
         $this->assertCount(0, $errors);
     }
+
+    public function testFromFailureWithMutateOperation()
+    {
+        $failureWithMutateOperation = new GoogleAdsFailure([
```
PierrickVoulet: Only thing that changes compared to the existing "failure" instance is the "field_name" value? It might be better factorizing them to avoid code duplication.

PierrickVoulet (changes requested):

tests/Google/Ads/GoogleAds/Util/V1/GoogleAdsErrorsTest.php
Raibaz: Right, done.

PierrickVoulet (approved):

Issue #191, opened 2019-08-15
Account Hierarchy Example


hackzilla:
I'm struggling to query our Account Hierarchy.

It seems to be that I need to query CustomerClient, then CustomerClientLink, then CustomerManagerLink and finally customer.

🤯

You had a great example in the previous library.
https://github.com/googleads/googleads-php-lib/blob/master/examples/AdWords/v201809/AccountManagement/GetAccountHierarchy.php

And I was hoping that the example ListAccessibleCustomers (#95) was the equivalent.

Any chance you could create an example?

fiboknacky:
Hello,

Thanks for your feedback. With the [recent release](https://developers.google.com/google-ads/api/docs/release-notes#account_management) that provides more info for the CustomerClient, the equivalent is now possible.

So, FYI, I'm now writing an example for this in #192. 

Best,
Knack

hackzilla:
I can see that this has been merged now.

PR #192, opened 2019-08-19
Added GetAccountHierarchy example.

fiboknacky:
This will also be a fix for #191.

### DIFF
--- a/examples/AccountManagement/GetAccountHierarchy.php
+++ b/examples/AccountManagement/GetAccountHierarchy.php
@@ -114,11 +114,6 @@ class GetAccountHierarchy
                 ['pageSize' => self::PAGE_SIZE]
             );
 
-            printf(
-                "Customer clients for the customer ID %d are found:%s",
-                $managerAccountId,
-                PHP_EOL
-            );
             // Iterates over all rows in all pages and prints the requested field values for
             // the change status in each row.
             foreach ($response->iterateAllElements() as $googleAdsRow) {
@@ -134,22 +129,11 @@ class GetAccountHierarchy
                     continue;
                 }
 
-                printf(
-                    "Customer client ID %d with name '%s', currency code '%s', time zone '%s'"
-                    . " is found.%s",
-                    $customerClient->getIdUnwrapped(),
-                    $customerClient->getDescriptiveNameUnwrapped(),
-                    $customerClient->getCurrencyCodeUnwrapped(),
-                    $customerClient->getTimeZoneUnwrapped(),
-                    PHP_EOL
-                );
-
                 // For all level-1 child accounts that are a manager account, the above query will
                 // be run against them to create an associative array of managers to their child
                 // accounts for printing the hierarchy afterwards.
                 $customerIdsToChildAccounts[$managerAccountId][] = $customerClient;
                 if ($customerClient->getManagerUnwrapped()) {
-                    print '  This customer account is also a manager.' . PHP_EOL;
                     // A customer can be managed by multiple managers, so to prevent visiting the
                     // same customer many times, we need to check if it's already in the map.
                     $alreadyVisited = array_key_exists(
@@ -161,10 +145,14 @@ class GetAccountHierarchy
                     }
                 }
             }
-            print PHP_EOL;
         };
 
-        print '(Customer ID, Descriptive Name)' . PHP_EOL;
+        printf(
+            "The hiearchy of the customer ID %d is printed below:%s",
+            $rootCustomerClient->getIdUnwrapped(),
+            PHP_EOL
+        );
+        print 'Customer ID (Descriptive Name, Currency Code, Time Zone)' . PHP_EOL;
         self::printAccountHierarchy($rootCustomerClient, $customerIdsToChildAccounts, 0);
     }
 
@@ -186,9 +174,11 @@ class GetAccountHierarchy
         $customerId = $customerClient->getIdUnwrapped();
         print str_repeat('-', $depth * 2);
         printf(
-            "%d, '%s'%s",
+            " %d ('%s', '%s', '%s')%s",
             $customerId,
             $customerClient->getDescriptiveNameUnwrapped(),
+            $customerClient->getCurrencyCodeUnwrapped(),
+            $customerClient->getTimeZoneUnwrapped(),
             PHP_EOL
         );
 
--- a/examples/AccountManagement/GetAccountHierarchy.php
+++ b/examples/AccountManagement/GetAccountHierarchy.php
@@ -32,11 +32,11 @@ use Google\Ads\GoogleAds\V2\Services\GoogleAdsRow;
 use Google\ApiCore\ApiException;
 
 /**
- * This example gets the account hierarchy of the specified customer account.
+ * This example gets the account hierarchy of the specified manager account.
  */
 class GetAccountHierarchy
 {
-    const CUSTOMER_ID = 'INSERT_CUSTOMER_ID_HERE';
+    const MANAGER_CUSTOMER_ID = 'INSERT_MANAGER_CUSTOMER_ID_HERE';
     const PAGE_SIZE = 1000;
 
     public static function main()
@@ -44,7 +44,7 @@ class GetAccountHierarchy
         // Either pass the required parameters for this example on the command line, or insert them
         // into the constants above.
         $options = (new ArgumentParser())->parseCommandArguments([
-            ArgumentNames::CUSTOMER_ID => GetOpt::REQUIRED_ARGUMENT
+            ArgumentNames::MANAGER_CUSTOMER_ID => GetOpt::REQUIRED_ARGUMENT
         ]);
 
         // Generate a refreshable OAuth2 credential for authentication.
@@ -59,7 +59,7 @@ class GetAccountHierarchy
         try {
             self::runExample(
                 $googleAdsClient,
-                $options[ArgumentNames::CUSTOMER_ID] ?: self::CUSTOMER_ID
+                $options[ArgumentNames::MANAGER_CUSTOMER_ID] ?: self::MANAGER_CUSTOMER_ID
             );
         } catch (GoogleAdsException $googleAdsException) {
             printf(
@@ -90,12 +90,13 @@ class GetAccountHierarchy
      * Runs the example.
      *
      * @param GoogleAdsClient $googleAdsClient the Google Ads API client
-     * @param int $customerId the customer ID
+     * @param int $managerCustomerId the customer ID
      */
-    public static function runExample(GoogleAdsClient $googleAdsClient, int $customerId)
+    public static function runExample(GoogleAdsClient $googleAdsClient, int $managerCustomerId)
     {
         $googleAdsServiceClient = $googleAdsClient->getGoogleAdsServiceClient();
-        // Creates a query that retrieves all child accounts of the specified customer ID.
+        // Creates a query that retrieves all child accounts of the manager specified in search
+        // calls below.
         $query = 'SELECT customer_client.client_customer, customer_client.level,'
             . ' customer_client.manager, customer_client.descriptive_name,'
             . ' customer_client.currency_code, customer_client.time_zone,'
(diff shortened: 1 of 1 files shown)


examples/AccountManagement/GetAccountHierarchy.php
```diff
@@ -0,0 +1,194 @@
...
+        while (!empty($unprocessedManagerAccounts)) {
+            $managerAccountId = array_shift($unprocessedManagerAccounts);
+            // Issues a search request by specifying page size.
+            $response = $googleAdsServiceClient->search(
+                $managerAccountId,
+                $query,
+                ['pageSize' => self::PAGE_SIZE]
+            );
+
+            // Iterates over all rows in all pages and prints the requested field values for
```
Raibaz: You may want to change this comment, as this loop doesn't print anything.

examples/AccountManagement/GetAccountHierarchy.php
```diff
@@ -0,0 +1,194 @@
...
+            // the change status in each row.
```
Raibaz: Also, it's not about change statuses

Raibaz (changes requested):

examples/AccountManagement/GetAccountHierarchy.php
```diff
@@ -0,0 +1,194 @@
...
+    {
+        $googleAdsServiceClient = $googleAdsClient->getGoogleAdsServiceClient();
+        // Creates a query that retrieves all child accounts of the specified customer ID.
+        $query = 'SELECT customer_client.client_customer, customer_client.level,'
+            . ' customer_client.manager, customer_client.descriptive_name,'
+            . ' customer_client.currency_code, customer_client.time_zone,'
+            . ' customer_client.id'
+            . ' FROM customer_client';
+
+        $unprocessedManagerAccounts = [$customerId];
```
PierrickVoulet: Add a comment to explain what we will do here, something like "Breadth-first search to gather all mappings of the hierarchy"

examples/AccountManagement/GetAccountHierarchy.php
```diff
@@ -0,0 +1,194 @@
...
+                $query,
+                ['pageSize' => self::PAGE_SIZE]
+            );
+
+            // Iterates over all rows in all pages and prints the requested field values for
+            // the change status in each row.
+            foreach ($response->iterateAllElements() as $googleAdsRow) {
+                /** @var GoogleAdsRow $googleAdsRow */
+                $customerClient = $googleAdsRow->getCustomerClient();
+                // The customer client that has level of 0 is the specified customer.
```
PierrickVoulet: Could be useful to elaborate a bit more  about the meaning of the level value

PierrickVoulet (changes requested):

examples/AccountManagement/GetAccountHierarchy.php
fiboknacky: Done.

examples/AccountManagement/GetAccountHierarchy.php
fiboknacky: Done.

Raibaz (approved):

PierrickVoulet (approved):

examples/AccountManagement/GetAccountHierarchy.php
```diff
@@ -0,0 +1,196 @@
...
+use Google\Ads\GoogleAds\V2\Resources\CustomerClient;
+use Google\Ads\GoogleAds\V2\Services\GoogleAdsRow;
+use Google\ApiCore\ApiException;
+
+/**
+ * This example gets the account hierarchy of the specified customer account.
+ */
+class GetAccountHierarchy
+{
+    const CUSTOMER_ID = 'INSERT_CUSTOMER_ID_HERE';
```
jradcliff: Let's clarify that this is a *manager* ID:
1. Call this `MANAGER_CUSTOMER_ID`
1. Change arg to runExample to `managerCustomerId`
1. Update the description on line 35 to say "of the specified manager account".

examples/AccountManagement/GetAccountHierarchy.php
```diff
@@ -0,0 +1,196 @@
...
+    /**
+     * Runs the example.
+     *
+     * @param GoogleAdsClient $googleAdsClient the Google Ads API client
+     * @param int $customerId the customer ID
+     */
+    public static function runExample(GoogleAdsClient $googleAdsClient, int $customerId)
+    {
+        $googleAdsServiceClient = $googleAdsClient->getGoogleAdsServiceClient();
+        // Creates a query that retrieves all child accounts of the specified customer ID.
```
jradcliff: "specified customer ID" could be confusing. I read it as meaning the $customerId param, but it's actually whichever manager you happen to be querying.

How about:
"Creates a query that retrieves all child accounts of the manager specified in search calls below"
or something similar?

examples/AccountManagement/GetAccountHierarchy.php
```diff
@@ -0,0 +1,196 @@
...
+                    );
+                    if (!$alreadyVisited && $customerClient->getLevelUnwrapped() === 1) {
+                        array_push($unprocessedManagerAccounts, $customerClient->getIdUnwrapped());
+                    }
+                }
+            }
+        };
+
+        printf(
+            "The hiearchy of the customer ID %d is printed below:%s",
```
jradcliff: "The hierarchy of manager customer ID %d is printed below:%s"

examples/AccountManagement/GetAccountHierarchy.php
```diff
@@ -0,0 +1,196 @@
...
+            $rootCustomerClient->getIdUnwrapped(),
+            PHP_EOL
+        );
+        print 'Customer ID (Descriptive Name, Currency Code, Time Zone)' . PHP_EOL;
```
jradcliff: nit: The format specified here is really dictated by the impl of `printAccountHierarchy()`. I'd move it there under a conditional of `if $depth === 0`

examples/AccountManagement/GetAccountHierarchy.php
```diff
@@ -0,0 +1,196 @@
...
+        }
+    }
+
+    /**
+     * Runs the example.
+     *
+     * @param GoogleAdsClient $googleAdsClient the Google Ads API client
+     * @param int $customerId the customer ID
+     */
+    public static function runExample(GoogleAdsClient $googleAdsClient, int $customerId)
```
jradcliff: @wihl had an excellent suggestion: if we make the manager customer ID optional, then this would be even more useful. If the user does not specify a manager customer ID, you could then call `listAccessibleCustomers` to get a list of managers accessible via the credentials, and then fetch and print the tree for each one of those managers.

fiboknacky: Let me confirm a bit more so I'm sure I understand correctly.
So, based on this suggestion we will:
- If no manager customer ID specified, call listAccessibleCustomers
- For each customer ID, call GoogleAdsService.Search() to check which ones are managers
- For all the managers, do what we originally do in this example -- getting all customer accounts under them.
?

examples/AccountManagement/GetAccountHierarchy.php
fiboknacky: SG.

examples/AccountManagement/GetAccountHierarchy.php
fiboknacky: And what happen if we have more than one root manager customer IDs?
Do we need to check if the accessible customers are actually child account of the other in the listAccessibleCustomers?

I find it's a bit complicated for the result printing process if we go down this path.

fiboknacky: Thanks for clarification. But that's actually a subset of what is possible.
In case we want to expand this example to use listaccessiblecustomers, we probably need to cover all possible cases.

In particular, what should we do when there are many customers accessible by the authenticated account, especially when some of them are child accounts of the others? Should we need to implement an algorithm to check the parent-child relationship? (That would further lengthen this example though) or should we just merely print the results for each accessible customers?

fiboknacky: Alright. Done.
Note that there are some places where I modified the variable to be just a customer ID because accessible customers can be non-manager ones, and I don't want to further complicate this example by trying to figure out whether each accessible customer is a manager or just a client.
This example in fact also works with non-manager customers too.

For the optional argument of the example, I retain it as manager customer ID though because it's more useful to use this example with a manager customer ID.

examples/AccountManagement/GetAccountHierarchy.php
```diff
@@ -0,0 +1,230 @@
...
+                        }
+                        continue;
+                    }
+
+                    // For all level-1 (direct child) accounts that are a manager account, the above
+                    // query will be run against them to create an associative array of managers to
+                    // their child accounts for printing the hierarchy afterwards.
+                    $customerIdsToChildAccounts[$customerId][] = $customerClient;
+                    if ($customerClient->getManagerUnwrapped()) {
+                        // A customer can be managed by multiple managers, so to prevent visiting
```
OrkhanAlikhanov: Is this comment really true? I am unable to create such hierarchy through GoogleAds UI.
I have one regular account `SubAccount` and 3 manager accounts: `RootManager, ManagerA, and ManagerB`
If I try to organize them like:
```
RootManager
>ManagerA
>>SubAccount
```
and 
```
ManagerB
>SubAccount
```

Everything is fine, I try to add `SubAccount` to `RootManager` and get error:
```
RootManager
>SubAccount /// ERROR: Invitee is already managed by another manager in your hierarchy
>ManagerA
>>SubAccount
```

If I try to add `ManagerB` to `RootManager`, I get error:
```
RootManager
>ManagerA
>>SubAccount
>ManagerB /// ERROR: The client would be linked more than once in a hierarchy
>>SubAccount
```

OrkhanAlikhanov: Now I realize that my account has access to to both `RootManager` and `ManagerB`, so `SubAccount` would show up twice. Nevermind, thank you!

examples/AccountManagement/GetAccountHierarchy.php
```diff
@@ -28,23 +28,29 @@
...
+use Google\Ads\GoogleAds\V2\Services\CustomerServiceClient;
 use Google\Ads\GoogleAds\V2\Services\GoogleAdsRow;
 use Google\ApiCore\ApiException;
 
 /**
- * This example gets the account hierarchy of the specified customer account.
+ * This example gets the account hierarchy of the specified manager account. If you don't specify
+ * manager customer ID, the example will instead print the hierarchies of all accessible customer
+ * accounts for your authenticated Google account.
+ * Note that if your accessible customers have manager-client relationships, you may get hierarchies
```
jradcliff: Instead of "have manager-client relationships...", please make it more general to cover the case where the user has direct access to a sub-manager as well, e.g.,

Note that if the list of accessible customers for your authenticated Google account includes accounts within the same hierarchy, this example will retrieve and print the overlapping portions of the hierarchy for each accessible customer.

jradcliff (approved):
Approved with one minor suggestion

examples/AccountManagement/GetAccountHierarchy.php
```diff
@@ -0,0 +1,231 @@
...
+        printf(
+            " %d ('%s', '%s', '%s')%s",
+            $customerId,
+            $customerClient->getDescriptiveNameUnwrapped(),
+            $customerClient->getCurrencyCodeUnwrapped(),
+            $customerClient->getTimeZoneUnwrapped(),
+            PHP_EOL
+        );
+
+        // Recursively call this function for all child accounts of $customerClient.
```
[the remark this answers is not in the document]
fiboknacky: I think it's quite obvious in that we store the array of customer ID to child accounts. Only a manager can has child accounts.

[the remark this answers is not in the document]
devchas: SGTM

PR #193, opened 2019-08-20
Fixed typehint of some nullable method variables of examples.

fiboknacky:


examples/AdvancedOperations/GetAdGroupBidModifiers.php
```diff
@@ -100,7 +100,7 @@ public static function main()
     public static function runExample(
         GoogleAdsClient $googleAdsClient,
         int $customerId,
-        int $adGroupId
+        ?int $adGroupId
```
PierrickVoulet: Comments are not updated in consequence: here it should be something like "int|null"

PierrickVoulet (changes requested):

PierrickVoulet (approved):

PR #195, opened 2019-08-22
Added AddFlightsFeed example.

fiboknacky:


examples/Remarketing/AddFlightsFeed.php
```diff
@@ -0,0 +1,423 @@
...
+     *      attributes
+     */
+    private static function createFeedMapping(
+        GoogleAdsClient $googleAdsClient,
+        int $customerId,
+        string $feedResourceName,
+        array $placeHoldersToFeedAttributesMap
+    ) {
+        // Maps the feed attribute IDs to the field ID constants.
+        $flightDescriptionMapping = new AttributeFieldMapping([
```
Raibaz: Could this be written as a loop over `$placeHoldersToFeedAttributesMap`?

Something like
```
$attributesMapping = [];
foreach($placeHoldersToFeedAttributesMap as $placeholder => $feedAttribute) {
    $attributesMapping []= new AttributeFieldMapping([ 
            'feed_attribute_id' => new Int64Value([ 'value' =>  $feedAttribute->getId()->getValue()]),
            'flight_field' => $placeholder
    ]);
}
```

Raibaz (changes requested):

examples/Remarketing/AddFlightsFeed.php
```diff
@@ -0,0 +1,423 @@
...
+     * Creates a flight feed.
+     *
+     * @param GoogleAdsClient $googleAdsClient the Google Ads API client
+     * @param int $customerId the customer ID
+     * @return string the feed resource name
+     */
+    private static function createFeed(GoogleAdsClient $googleAdsClient, int $customerId)
+    {
+        // Creates a flight description attribute.
+        $flightDescriptionAttribute = new FeedAttribute([
```
PierrickVoulet: All these attribute creations could be factorized into a "createAttribute" method directly called  from the feed object creation

examples/Remarketing/AddFlightsFeed.php
```diff
@@ -0,0 +1,423 @@
...
+            'type' =>  FeedAttributeType::STRING,
+            'name' => new StringValue(['value' => 'Flight Sale Price'])
+        ]);
+        // Creates a Final URLs attribute.
+        $finalUrlsAttribute = new FeedAttribute([
+            'type' =>  FeedAttributeType::URL_LIST,
+            'name' => new StringValue(['value' => 'Final URLs'])
+        ]);
+
+        // Creates the feed.
```
PierrickVoulet: nit: "... with the newly created attributes"

examples/Remarketing/AddFlightsFeed.php
```diff
@@ -0,0 +1,423 @@
...
+        $feedResourceName = $response->getResults()[0]->getResourceName();
+        printf("Feed with resource name '%s' was created.%s", $feedResourceName, PHP_EOL);
+
+        return $feedResourceName;
+    }
+
+    /**
+     * Retrieves details about a feed. The initial query retrieves the feed attributes, or columns,
+     * of the feed. Each feed attribute will also include the feed attribute ID, which will be used
+     * in a subsequent step. The example then inserts a new key, value pair into a map for each
```
PierrickVoulet: "key-value" instead of "key, value"

PierrickVoulet: I would breakdown into 2 paragraphs with the second starting at "The example...". In the second  paragraph I would use bullet points with facts to make it easier to understand.

examples/Remarketing/AddFlightsFeed.php
```diff
@@ -0,0 +1,423 @@
...
+     * feed attribute, which is the return value of the method. The keys are the placeholder types
+     * that the columns will be. The values are the feed attributes.
+     *
+     * @param GoogleAdsClient $googleAdsClient the Google Ads API client
+     * @param int $customerId the customer ID
+     * @param string $feedResourceName the feed resource name to get its attributes
```
PierrickVoulet: nit: "to get the attributes from"

examples/Remarketing/AddFlightsFeed.php
```diff
@@ -0,0 +1,423 @@
...
+        GoogleAdsClient $googleAdsClient,
+        int $customerId,
+        string $feedResourceName
+    ) {
+        $googleAdsServiceClient = $googleAdsClient->getGoogleAdsServiceClient();
+        // Constructs the query to get the feed attributes for the specified feed resource name.
+        $query = "SELECT feed.attributes FROM feed WHERE feed.resource_name = '$feedResourceName'";
+        // Issues a search request by specifying page size.
+        $response =
+            $googleAdsServiceClient->search($customerId, $query, ['pageSize' => self::PAGE_SIZE]);
```
PierrickVoulet: Why  not using a page size of 1 instead?

examples/Remarketing/AddFlightsFeed.php
PierrickVoulet: All these attribute field mapping creations could be factorized into a "createAttributeFieldMapping" method directly called from the feedMapping object creation

examples/Remarketing/AddFlightsFeed.php
```diff
@@ -0,0 +1,423 @@
...
+     *      attributes
+     */
+    private static function createFeedItem(
+        GoogleAdsClient $googleAdsClient,
+        int $customerId,
+        string $feedResourceName,
+        array $placeHoldersToFeedAttributesMap
+    ) {
+        // Creates the flight description feed attribute value.
+        $flightDescriptionAttributeValue = new FeedItemAttributeValue([
```
PierrickVoulet: All these feed item attribute value creations could be factorized into a "createFeedItemAttributeValue" method directly called from the feedItem object creation

PierrickVoulet (changes requested):

examples/Remarketing/AddFlightsFeed.php
```diff
@@ -0,0 +1,423 @@
...
+        self::createFeedItem(
+            $googleAdsClient,
+            $customerId,
+            $feedResourceName,
+            $placeHoldersToFeedAttributesMap
+        );
+    }
+
+    /**
+     * Creates a flight feed.
```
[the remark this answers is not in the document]
fiboknacky: That's fair but this example is about creating a feed for a flight (which technically is what you described here).

I'm open for any specific suggestion, but I don't think there is a concise way to describe this unless we want to go for long version.
If we're sensitive with this, we probably need to modify the file's PHPDoc as well.

I can just drop a "flight" word if that bugs you. :)

examples/Remarketing/AddFlightsFeed.php
fiboknacky: Sorry, don't get what you mean.

examples/Remarketing/AddFlightsFeed.php
fiboknacky: Good question. This is done in the golden PR. I guess it's just copied over from other examples.
A bigger question is whether we want to micromanage this that much. 
Large page size should cause no harm in this case, no?

One small advantage I see is when people copy this code to their application , they don't need to worry much about too small page size.
In all examples, we use this same style over and over, so people are already used to it. They can just expect that they can configure PAGE_SIZE by their owns after copying the code.

examples/Remarketing/AddFlightsFeed.php
fiboknacky: Do you mean creating a private method for that?
I'm rather not nit-picky on this so we don't diverge much from the golden PR.
The current block is small enough to fit here with reasonable variable names used.
The part where all these variables are passed to FeedMapping creation also has reasonable readability as well.

examples/Remarketing/AddFlightsFeed.php
fiboknacky: Likewise, if you mean creating another private method, I'd rather not do it this time because it would cause some complication when we write a guide from these code examples. 
The current code, though not super efficient, I found the readability level is still reasonable.

examples/Remarketing/AddFlightsFeed.php
fiboknacky: That's probably possible too.

But I'd rather follow the golden PR here because it clearly shows the steps of how fields and objects are needed to be created for feed, feed mapping and feed items.
The three private methods are parallel to each other and have the exact same structure, which could help people follow easily.

Besides, stating the FlightPlaceholderField::XXXX explicitly serves an educational purpose to repeatedly show readers what placeholder fields belong to the flight feed.

examples/Remarketing/AddFlightsFeed.php
PierrickVoulet: Yes this is what I meant and I agree with your points.

examples/Remarketing/AddFlightsFeed.php
PierrickVoulet: If to request more than expected is not considered as a bad practice then I believe it is good to keep it like that for the reasons you highlighted.

PierrickVoulet (approved):

examples/Remarketing/AddFlightsFeed.php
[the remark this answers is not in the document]
Raibaz: How about "creates a feed to be used as a flight feed"? Too long?

Raibaz (approved):
LGTM with a minor nit

PR #213, opened 2019-09-25
Added examples of updating and removing flights feed item string attribute values

fiboknacky:
* Some parts of AddFlightsFeed are extracted to be stored in Feeds Utils, so they can be shared in other examples.
* placeholderFieldsMapFor() is designed to be generic. That's why there is another similar function named flightPlaceholderFieldsMapFor(). In the future, if we need to create a similar example for real estate, for example, we can just supply a new callable to placeholderFieldsMapFor().

### DIFF
--- a/examples/Remarketing/AddFlightsFeed.php
+++ b/examples/Remarketing/AddFlightsFeed.php
@@ -22,6 +22,7 @@ require __DIR__ . '/../../vendor/autoload.php';
 use GetOpt\GetOpt;
 use Google\Ads\GoogleAds\Examples\Utils\ArgumentNames;
 use Google\Ads\GoogleAds\Examples\Utils\ArgumentParser;
+use Google\Ads\GoogleAds\Examples\Utils\Feeds;
 use Google\Ads\GoogleAds\Lib\OAuth2TokenBuilder;
 use Google\Ads\GoogleAds\Lib\V2\GoogleAdsClient;
 use Google\Ads\GoogleAds\Lib\V2\GoogleAdsClientBuilder;
@@ -39,7 +40,6 @@ use Google\Ads\GoogleAds\V2\Resources\FeedMapping;
 use Google\Ads\GoogleAds\V2\Services\FeedItemOperation;
 use Google\Ads\GoogleAds\V2\Services\FeedMappingOperation;
 use Google\Ads\GoogleAds\V2\Services\FeedOperation;
-use Google\Ads\GoogleAds\V2\Services\GoogleAdsRow;
 use Google\ApiCore\ApiException;
 use Google\Protobuf\Int64Value;
 use Google\Protobuf\StringValue;
@@ -111,7 +111,7 @@ class AddFlightsFeed
         // Gets the newly created feed's attributes and packages them into a map. This read
         // operation is required to retrieve the attribute IDs.
         $placeHoldersToFeedAttributesMap =
-            self::getFeed($googleAdsClient, $customerId, $feedResourceName);
+            Feeds::flightPlaceholderFieldsMapFor($feedResourceName, $customerId, $googleAdsClient);
         // Creates the feed mapping.
         self::createFeedMapping(
             $googleAdsClient,
@@ -189,68 +189,6 @@ class AddFlightsFeed
     }
 
     /**
-     * Retrieves details about a feed. The initial query retrieves the feed attributes, or columns,
-     * of the feed. Each feed attribute will also include the feed attribute ID, which will be used
-     * in a subsequent step.
-     *
-     * The example then inserts a new key-value pair into a map for each
-     * feed attribute, which is the return value of the method:
-     * - The keys are the placeholder types that the columns will be.
-     * - The values are the feed attributes.
-     *
-     * @param GoogleAdsClient $googleAdsClient the Google Ads API client
-     * @param int $customerId the customer ID
-     * @param string $feedResourceName the feed resource name to get the attributes from
-     * @return array the map from placeholder fields to feed attributes
-     */
-    private static function getFeed(
-        GoogleAdsClient $googleAdsClient,
-        int $customerId,
-        string $feedResourceName
-    ) {
-        $googleAdsServiceClient = $googleAdsClient->getGoogleAdsServiceClient();
-        // Constructs the query to get the feed attributes for the specified feed resource name.
-        $query = "SELECT feed.attributes FROM feed WHERE feed.resource_name = '$feedResourceName'";
-        // Issues a search request by specifying page size.
-        $response =
-            $googleAdsServiceClient->search($customerId, $query, ['pageSize' => self::PAGE_SIZE]);
-
-        // Gets the first result because we only need the single feed we created previously.
-        /** @var GoogleAdsRow $googleAdsRow */
-        $googleAdsRow = $response->getIterator()->current();
-
-        // Gets the attributes list from the feed and creates a map with keys of each attribute and
-        // values of each corresponding ID.
-        $feedAttributes =
-            iterator_to_array($googleAdsRow->getFeed()->getAttributes()->getIterator());
-
-        $feedPlaceHolderFieldsList = array_map(function (FeedAttribute $feedAttribute) {
-            switch ($feedAttribute->getNameUnwrapped()) {
-                case 'Flight Description':
-                    return FlightPlaceholderField::FLIGHT_DESCRIPTION;
-                    break;
-                case 'Destination ID':
-                    return FlightPlaceholderField::DESTINATION_ID;
-                    break;
-                case 'Flight Price':
-                    return FlightPlaceholderField::FLIGHT_PRICE;
-                    break;
-                case 'Flight Sale Price':
-                    return FlightPlaceholderField::FLIGHT_SALE_PRICE;
-                    break;
-                case 'Final URLs':
-                    return FlightPlaceholderField::FINAL_URLS;
-                    break;
-                // See FlightPlaceholderField.php for all available values.
-                default:
-                    throw new \RuntimeException('Invalid feed attribute name.');
-            }
-        }, $feedAttributes);
-
-        return array_combine($feedPlaceHolderFieldsList, $feedAttributes);
-    }
-
-    /**
      * Creates a feed mapping for a given feed.
      *
      * @param GoogleAdsClient $googleAdsClient the Google Ads API client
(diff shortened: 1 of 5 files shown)


examples/Utils/Feeds.php
```diff
@@ -0,0 +1,178 @@
...
+
+    /**
+     * Gets the index of the target feed item attribute value. This is needed to specify which feed
+     * item attribute value will be updated in the given feed item.
+     *
+     * @param FeedItemAttributeValue $targetFeedItemAttributeValue the new feed item attribute value
+     *     that will be updated
+     * @param FeedItem $feedItem the feed item that will be updated
+     */
+    public static function attributeIndexFor(
```
PierrickVoulet: nit:I would name it 'getAttributeValueIndexOfFeedItem' and simplify the comments and variable names

examples/Utils/Feeds.php
```diff
@@ -0,0 +1,178 @@
...
+
+    /**
+     * Retrieves a feed item and its attribute values given a resource name.
+     *
+     * @param string $feedItemResourceName the feed item resource name
+     * @param int $customerId the customer ID
+     * @param GoogleAdsClient $googleAdsClient the Google Ads API client
+     * @return FeedItem the feed item
+     */
+    public static function feedItemFor(
```
PierrickVoulet: nit:I would name it 'getFeedItemWithAttributeValues'

examples/Utils/Feeds.php
```diff
@@ -0,0 +1,178 @@
...
+                'No matching feed attribute for feed item attribute ID: '
+                . $targetFeedItemAttributeValue->getFeedAttributeIdUnwrapped()
+            );
+        }
+
+        return $attributeIndex;
+    }
+
+    /**
+     * Retrieves the place holder fields to field attributes map for a flights feed.
```
PierrickVoulet: to '**feed** attributes' instead of '**field** attributes'

examples/Utils/Feeds.php
```diff
@@ -0,0 +1,178 @@
...
+    public static function flightPlaceholderFieldsMapFor(
+        string $feedResourceName,
+        int $customerId,
+        GoogleAdsClient $googleAdsClient
+    ) {
+        return self::placeholderFieldsMapFor(
+            $feedResourceName,
+            $customerId,
+            $googleAdsClient,
+            function (FeedAttribute $feedAttribute) {
```
PierrickVoulet: Why not providing a map here instead of mapping function as a parameter?

The value here would be something like:
[
'Flight Description' => FlightPlaceholderField::FLIGHT_DESCRIPTION,
'Destination ID' => FlightPlaceholderField::DESTINATION_ID,
...
]

And a generic function that leverages this mapping parameter would be implemented for all in the 'placeholderFieldsMapFor' method

examples/Utils/Feeds.php
```diff
@@ -0,0 +1,178 @@
...
+            }
+        );
+    }
+
+    /**
+     * Retrieves the place holder fields to field attributes map for a feed. The initial
+     * query retrieves the feed attributes, or columns, of the feed. Each feed attribute will also
+     * include the feed attribute ID, which will be used in a subsequent step.
+     *
+     * The example then inserts a new key-value pair into a map for each
```
PierrickVoulet: This explains example logic in an utility class: we should relocate this elsewhere even if it has to be duplicated in each example.  This applies to all comments in the utility class

examples/Utils/Feeds.php
```diff
@@ -0,0 +1,178 @@
...
+     * feed attribute, which is the return value of the method:
+     * - The keys are the placeholder types that the columns will be.
+     * - The values are the feed attributes.
+     *
+     * @param string $feedResourceName the feed resource name to get the attributes from
+     * @param int $customerId the customer ID
+     * @param GoogleAdsClient $googleAdsClient the Google Ads API client
+     * @param callable $mappingFunction the function that maps feed attribute names to placeholder
```
PierrickVoulet: Change for a map, related to another comment above

examples/Remarketing/RemoveFlightsFeedItemStringAttributeValue.php
```diff
@@ -0,0 +1,174 @@
...
+        // Gets a map of the placeholder values to feed attributes.
+        $placeHoldersToFeedAttributesMap = Feeds::flightPlaceholderFieldsMapFor(
+            ResourceNames::forFeed($customerId, $feedId),
+            $customerId,
+            $googleAdsClient
+        );
+        // Gets the ID of the feed attribute for the placeholder field.
+        $attributeId = $placeHoldersToFeedAttributesMap[
+            FlightPlaceholderField::value($flightPlaceholderFieldName)]->getIdUnwrapped();
+        // Creates the feed item attribute value that will be removed.
```
PierrickVoulet: Add a comment to explain that only the ID is required here

PierrickVoulet (changes requested):

examples/Remarketing/AddFlightsFeed.php
```diff
@@ -111,7 +112,7 @@ public static function runExample(GoogleAdsClient $googleAdsClient, int $custome
         // Gets the newly created feed's attributes and packages them into a map. This read
         // operation is required to retrieve the attribute IDs.
         $placeHoldersToFeedAttributesMap =
-            self::getFeed($googleAdsClient, $customerId, $feedResourceName);
+            Feeds::flightPlaceholderFieldsMapFor($feedResourceName, $customerId, $googleAdsClient);
```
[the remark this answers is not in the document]
fiboknacky: I deliberately shuffle it like this.
The name of the method "flightPlaceholderFieldsMapFor" will be more meaningful when we place the feed resource name as the first parameter: "flight placeholder fields map for a feed resource name"

Consistency with other code examples would be good too, but this one is not the method of the code example.
It belongs to the utils.

I can do flightPlaceholderFieldsMapFor($feedResourceName, $googleAdsClient, $customerId) as well.

examples/Utils/Feeds.php
fiboknacky: The util in this library tries to follow Java's guava like ImmutableSet.of().
That's why I rarely use getXXXXofYYYY format in this util.

examples/Utils/Feeds.php
fiboknacky: That's because I need to use array_map which requires a callable function.
It doesn't require an associative array.
Or do you have other ways around to propose for the generic function too?

examples/Utils/Feeds.php
fiboknacky: Same reason with the other threads. This is intentional.
See reference in other utils like ArgumentNames, GoogleAdsErrors, EnvironmentalVariables.

examples/Utils/Feeds.php
fiboknacky: See my comment above. I don't think using a map is more convenient here.
At the end of this method, the callable is needed.

examples/Remarketing/AddFlightsFeed.php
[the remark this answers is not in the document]
PierrickVoulet: Got it, I think it is nit so feel free to keep it as it is.

examples/Utils/Feeds.php
PierrickVoulet: I think you could make it work with the `use` keyword to introduce the associative array as a local scope variable of the callable function and rework this last to be generic. Something like that:

```php
$placeHolderFieldMappings = [
    'Flight Description' => FlightPlaceholderField::FLIGHT_DESCRIPTION,
    'Destination ID' => FlightPlaceholderField::DESTINATION_ID,
    ...
]

$callable = function (FeedAttribute $feedAttribute) use ($placeHolderFieldMappings) {
    // TODO: return RuntimeException if the given feed attribute is undefined in the mappings
    return $placeHolderFieldMappings[$feedAttribute];
}

return self::placeholderFieldsMapFor(
            $feedResourceName,
            $customerId,
            $googleAdsClient,
            $callable
);
```

WDYT?

fiboknacky: That looks great to me too so I just changed to it.

examples/Remarketing/UpdateFlightsFeedItemStringAttributeValue.php
```diff
@@ -0,0 +1,179 @@
...
+        $feedItem = Feeds::feedItemFor(
+            ResourceNames::forFeedItem($customerId, $feedId, $feedItemId),
+            $customerId,
+            $googleAdsClient
+        );
+        // Gets the index of the attribute value that will be updated in the feed item.
+        $attributeIndex = Feeds::attributeIndexFor($updatedFeedItemAttributeValue, $feedItem);
+        // Any feed item attribute values that are not included in the updated feed item will be
+        // removed from the feed item, which is why you must create the feed item from the existing
+        // feed item and its attribute values. Then, update only the field that you want.
```
jradcliff: nit: "field" -> "attribute"

examples/Utils/ArgumentNames.php
```diff
@@ -81,6 +86,9 @@ final class ArgumentNames
         self::CRITERION_ID => 'The criterion ID',
         self::CUSTOMER_ID => 'The customer ID without dashes',
         self::DRAFT_ID => 'The draft ID',
+        self::FEED_ID => 'The feed ID',
+        self::FEED_ITEM_ID => 'The feed item ID',
+        self::FLIGHT_PLACEHOLDER_FIELD_NAME => 'The flight placeholder field name',
```
jradcliff: WDYT of just naming this FEED_PLACEHOLDER_FIELD_NAME? Then you can reuse it in other feed examples later.

examples/Utils/Feeds.php
```diff
@@ -0,0 +1,174 @@
...
+        return $response->getIterator()->current()->getFeedItem();
+    }
+
+    /**
+     * Gets the index of the target feed item attribute value. This is needed to specify which feed
+     * item attribute value will be updated in the given feed item.
+     *
+     * @param FeedItemAttributeValue $targetFeedItemAttributeValue the new feed item attribute value
+     *     that will be updated
+     * @param FeedItem $feedItem the feed item that will be updated
```
jradcliff: It's worth mentioning here that the caller should pass a `FeedItem` here with the `attributeValues` populated

examples/Utils/Feeds.php
```diff
@@ -0,0 +1,174 @@
...
+                'No matching feed attribute for feed item attribute ID: '
+                . $targetFeedItemAttributeValue->getFeedAttributeIdUnwrapped()
+            );
+        }
+
+        return $attributeIndex;
+    }
+
+    /**
+     * Retrieves the place holder fields to feed attributes map for a flights feed.
```
jradcliff: Add a note about name assumptions, e.g., "Assumes that the field names of the feed use the standard names listed \[here\](http://link-to-help-center-article-if-it-exists)."

PierrickVoulet (approved):

examples/Utils/Feeds.php
fiboknacky: I moved the below comment here instead.
This should be equivalent to linking to external articles.

jradcliff (approved):

PR #214, opened 2019-09-26
Changes for release v2_2.

PierrickVoulet:


src/Google/Ads/GoogleAds/V2/Services/resources/mutate_job_service_rest_client_config.php
```diff
@@ -62,64 +62,5 @@
                 ],
             ],
         ],
-        'google.longrunning.Operations' => [
```
fiboknacky: Would this affect mutate job service? It's currently using longrunning.
Could you follow up after the release?

fiboknacky (approved):

src/Google/Ads/GoogleAds/V2/Services/resources/mutate_job_service_rest_client_config.php
PierrickVoulet: Good catch, I will follow up on this.

PR #225, opened 2019-11-01
Add get product bidding category constant

PierrickVoulet:


examples/ShoppingAds/GetProductBiddingCategoryConstant.php
```diff
@@ -0,0 +1,179 @@
...
+
+            if ($productBiddingCategory->getProductBiddingCategoryConstantParent() != null) {
+                $parentResourceName =
+                    $productBiddingCategory->getProductBiddingCategoryConstantParentUnwrapped();
+                if (!array_key_exists($parentResourceName, $biddingCategories)) {
+                    $biddingCategories[$parentResourceName] = [];
+                }
+                $biddingCategories[$parentResourceName]['children'][$resourceName] = $biddingCategories[$resourceName];
+            } else {
+                $rootCategories[$resourceName] = $biddingCategories[$resourceName];
```
devchas: Since arrays in PHP aren't passed by reference, if a child bidding category is encountered after its parent bidding category is assigned to the rootCategories array, would that child be reflected in the final rootCategories array?

examples/ShoppingAds/GetProductBiddingCategoryConstant.php
```diff
@@ -0,0 +1,179 @@
...
+use Google\Ads\GoogleAds\Lib\V2\GoogleAdsClient;
+use Google\Ads\GoogleAds\Lib\V2\GoogleAdsClientBuilder;
+use Google\Ads\GoogleAds\Lib\V2\GoogleAdsException;
+use Google\Ads\GoogleAds\Lib\OAuth2TokenBuilder;
+use Google\Ads\GoogleAds\V2\Errors\GoogleAdsError;
+use Google\Ads\GoogleAds\V2\Resources\ProductBiddingCategoryConstant;
+use Google\Ads\GoogleAds\V2\Services\GoogleAdsRow;
+use Google\ApiCore\ApiException;
+
+/** Fetches the set of valid ProductBiddingCategories.*/
```
devchas: ...and prints the results recursively.

PierrickVoulet: Using the same  sentence as the [golden](https://github.com/googleads/google-ads-java/blob/master/google-ads-examples/src/main/java/com/google/ads/googleads/examples/shoppingads/GetProductBiddingCategoryConstant.java#L36)

examples/ShoppingAds/GetProductBiddingCategoryConstant.php
PierrickVoulet: Good one, tx

examples/ShoppingAds/GetProductBiddingCategoryConstant.php
```diff
@@ -0,0 +1,186 @@
...
+            product_bidding_category_constant.product_bidding_category_constant_parent
+            FROM product_bidding_category_constant
+            WHERE product_bidding_category_constant.country_code IN (\'US\')
+            ORDER BY product_bidding_category_constant.localized_name ASC';
+
+        // Creates a map of top level categories.
+        $rootCategories = [];
+        // Creates a map of all categories found in the results.
+        // This is a convenience lookup to enable fast retrieval of existing categories.
+        $biddingCategories = [];
```
fiboknacky: This could be moved to below "search()".
They're not used in the search and this move will make them close to the for-loop, where they're used.

examples/ShoppingAds/GetProductBiddingCategoryConstant.php
```diff
@@ -0,0 +1,186 @@
...
+                $biddingCategories[$parentResourceName]['children'][$resourceName] =
+                    &$biddingCategories[$resourceName];
+            } else {
+                // Otherwise adds the category as a root category.
+                $rootCategories[$resourceName] = &$biddingCategories[$resourceName];
+            }
+        }
+
+        // Prints the result.
+        self::displayCategories($rootCategories, "");
```
fiboknacky: Nit: As you tried to use '' everywhere, let's use '' instead of "" here too.

examples/ShoppingAds/GetProductBiddingCategoryConstant.php
```diff
@@ -0,0 +1,186 @@
...
+                $biddingCategories[$resourceName] = [];
+            }
+            // Ensures that the localized name attribute for the categories is set.
+            // It will not be initialized for new categories or categories added
+            // to biddingCategories as a result of being a parent category below.
+            $biddingCategories[$resourceName]['localizedName'] =
+                $productBiddingCategory->getLocalizedNameUnwrapped();
+
+            if ($productBiddingCategory->getProductBiddingCategoryConstantParent() != null) {
+                // If the category has a parent.
```
fiboknacky: Nit: This is not quite useful comment. It just explains the if clause as is.
Consider explaining more about "why" or "how" rather than "what" here.

examples/ShoppingAds/GetProductBiddingCategoryConstant.php
```diff
@@ -0,0 +1,186 @@
...
+                // Adds the parent category in the map if new.
+                if (!array_key_exists($parentResourceName, $biddingCategories)) {
+                    $biddingCategories[$parentResourceName] = [];
+                }
+                // Adds the category as a child category of the parent category.
+                $biddingCategories[$parentResourceName]['children'][$resourceName] =
+                    &$biddingCategories[$resourceName];
+            } else {
+                // Otherwise adds the category as a root category.
+                $rootCategories[$resourceName] = &$biddingCategories[$resourceName];
```
fiboknacky: Is the reference access (&) here needed? I've done some simply mapping in other examples before. It didn't seem we need it.

examples/ShoppingAds/GetProductBiddingCategoryConstant.php
```diff
@@ -0,0 +1,186 @@
...
+            $productBiddingCategory = $googleAdsRow->getProductBiddingCategoryConstant();
+            $resourceName = $productBiddingCategory->getResourceName();
+
+            // Adds the category in the map if new.
+            if (!array_key_exists($resourceName, $biddingCategories)) {
+                $biddingCategories[$resourceName] = [];
+            }
+            // Ensures that the localized name attribute for the categories is set.
+            // It will not be initialized for new categories or categories added
+            // to biddingCategories as a result of being a parent category below.
```
fiboknacky: Overall, I understand now what it means, but it was confusing at first because we don't see the below block yet.

Besides, if the $resourceName is in the $biddingCategories already, does it mean it has been added by the "parent detection" below?
if so, can't we just put the line 133 into the if clause above?
That way, we can probably do something like this:

```php
// Adds the category to the map if it's new. Some categories might be added before as a result of being the 
// parent of another categories. See the code below for detail.
if (!array_key_exists($resourceName, $biddingCategories)) {
    $biddingCategories[$resourceName] = [];
    $biddingCategories[$resourceName]['localizedName'] =
        $productBiddingCategory->getLocalizedNameUnwrapped();
}
           
```

Note that our comment is shorter and easier to grasp.

examples/ShoppingAds/GetProductBiddingCategoryConstant.php
```diff
@@ -0,0 +1,186 @@
...
+            $biddingCategories[$resourceName]['localizedName'] =
+                $productBiddingCategory->getLocalizedNameUnwrapped();
+
+            if ($productBiddingCategory->getProductBiddingCategoryConstantParent() != null) {
```
fiboknacky: Nit: use !is_null or !== for consistency

examples/ShoppingAds/GetProductBiddingCategoryConstant.php
PierrickVoulet: I had to add them both to make it work as intended (one line printed per product bidding category constant). If I do not use references, only corner cases make it through (~500/7000).

examples/ShoppingAds/GetProductBiddingCategoryConstant.php
PierrickVoulet: I like your idea of set the value only when necessary. To avoid confusion, I wrapped the statement in a basic condition which makes it less confusing. LMK what you think

PR #236, opened 2019-11-20
Fixing AdGroupAd status to be ENABLED for Hotels

stevenberger:
There was a change for Hotels that causes an error when setting AdGroupAd to PAUSED and should always be ENABLED.

### DIFF
--- a/examples/HotelAds/AddHotelAd.php
+++ b/examples/HotelAds/AddHotelAd.php
@@ -328,7 +328,7 @@ class AddHotelAds
         // Creates a new ad group ad and sets the hotel ad to it.
         $adGroupAd = new AdGroupAd([
             'ad' => $ad,
-            'status' => AdGroupAdStatus::PAUSED,
+            'status' => AdGroupAdStatus::ENABLED,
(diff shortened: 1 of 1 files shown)


fiboknacky:
Could you elaborate what kind of error it is?

stevenberger:
Setting `AdGroupAd`'s status to `PAUSED` will now throw a `OPERATION_NOT_PERMITTED_FOR_CAMPAIGN_TYPE`, this is expected as the status for `AdGroupAd` for Hotel campaigns should always be `ENABLED`.

fiboknacky:
OK. I'll coordinate across other libraries.

stevenberger:
@fiboknacky I actually have a PR open with all the other libraries (Java and PHP already merged).

[Python](https://github.com/googleads/google-ads-python/pull/186)
[.NET](https://github.com/googleads/google-ads-dotnet/pull/123)
[Ruby](https://github.com/googleads/google-ads-ruby/pull/192)

fiboknacky:
OK. Thanks!

PR #238, opened 2019-11-25
Added AddDemoGraphicTargetingCriteria example.

fiboknacky:


examples/Targeting/AddDemographicTargetingCriteria.php
```diff
@@ -0,0 +1,159 @@
...
+use Google\Ads\GoogleAds\V2\Errors\GoogleAdsError;
+use Google\Ads\GoogleAds\V2\Resources\AdGroupCriterion;
+use Google\Ads\GoogleAds\V2\Services\AdGroupCriterionOperation;
+use Google\ApiCore\ApiException;
+use Google\Protobuf\BoolValue;
+use Google\Protobuf\StringValue;
+
+/**
+ * This example adds demographic target criteria to an ad group, one as positive ad group criterion
+ * and one as a negative ad group criterion. To get ad groups, run AddAdGroups.php.
```
PierrickVoulet: nit: either "one as **a** positive" or "one as ~~a~~ negative"

PierrickVoulet: nit: "To create ad groups" instead of "To get ad groups"

examples/Targeting/AddDemographicTargetingCriteria.php
```diff
@@ -0,0 +1,159 @@
...
+     * @param int $customerId the client customer ID
+     * @param int $adGroupId the ad group ID to add demographic targeting criteria to
+     */
+    public static function runExample(
+        GoogleAdsClient $googleAdsClient,
+        int $customerId,
+        int $adGroupId
+    ) {
+        $adGroupResourceName = new StringValue(
+            ['value' => ResourceNames::forAdGroup($customerId, $adGroupId)]
```
PierrickVoulet: nit: keep the "[" and "]" next to "(" and ")"

examples/Targeting/AddDemographicTargetingCriteria.php
```diff
@@ -0,0 +1,159 @@
...
+     * @param GoogleAdsClient $googleAdsClient the Google Ads API client
+     * @param int $customerId the client customer ID
+     * @param int $adGroupId the ad group ID to add demographic targeting criteria to
+     */
+    public static function runExample(
+        GoogleAdsClient $googleAdsClient,
+        int $customerId,
+        int $adGroupId
+    ) {
+        $adGroupResourceName = new StringValue(
```
PierrickVoulet: nit: would add a comment

examples/Targeting/AddDemographicTargetingCriteria.php
```diff
@@ -0,0 +1,159 @@
...
+            ['value' => ResourceNames::forAdGroup($customerId, $adGroupId)]
+        );
+
+        // Creates an ad group criterion for gender.
```
PierrickVoulet: nit: would explicitly say it is a "positive" criterion

PierrickVoulet (approved):
I have only nits so approved

devchas (approved):

examples/Targeting/AddDemographicTargetingCriteria.php
fiboknacky: I deliberately not do this because 
* The code is quite so obvious what it does.
* It's just a small part of bigger main steps: create ad group -> create ad group operation -> issue a mutate request on server
* It's just the portion that was extracted from the code below. As we need to do this twice, it's pulled out and made local variable. Originally, there was no code comment for this too.

PR #242, opened 2019-11-28
Add upload offline conversions

PierrickVoulet:


fiboknacky:
Let me request Nadine instead, as Josh has been requested in so many PRs right now.

examples/Remarketing/UploadOfflineConversions.php
```diff
@@ -0,0 +1,157 @@
...
+ * This code example imports offline conversion values for specific clicks to
+ * your account. To get Google Click ID for a click, use the "click_view"
+ * resource of report. To set up a conversion action, run the
+ * AddConversionAction.php example.
+ */
+class UploadOfflineConversions
+{
+    const CUSTOMER_ID = 'INSERT_CUSTOMER_ID_HERE';
+    const CONVERSION_ACTION_ID = 'INSERT_CONVERSION_ACTION_ID_HERE';
+    const GCL_ID = 'INSERT_GCL_ID_HERE';
```
fiboknacky: Based on several sources, it seems the most frequently used abbreviation is just "gclid", not "gclId".
I know we were not consistent in the past, but let's fix this here by using "gclid".
So, this const should be named GCLID, since gcl and id are in the same word.

examples/Remarketing/UploadOfflineConversions.php
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+     * Runs the example.
+     *
+     * @param GoogleAdsClient $googleAdsClient the Google Ads API client
+     * @param int $customerId the customer ID
+     * @param int $conversionActionId the ID of the conversion action to upload to
+     * @param string $gclId the GCLID for the conversion (should be newer than 30 days)
+     * @param string $conversionTime the date and time of the conversion (should be after the
+     * click time)
```
fiboknacky: Please indent this line.

examples/Remarketing/UploadOfflineConversions.php
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    /**
+     * Runs the example.
+     *
+     * @param GoogleAdsClient $googleAdsClient the Google Ads API client
+     * @param int $customerId the customer ID
+     * @param int $conversionActionId the ID of the conversion action to upload to
+     * @param string $gclId the GCLID for the conversion (should be newer than 30 days)
+     * @param string $conversionTime the date and time of the conversion (should be after the
```
fiboknacky: Are there any links to format of datetime that we can use?

examples/Remarketing/UploadOfflineConversions.php
```diff
@@ -0,0 +1,157 @@
...
+     * click time)
+     * @param float $conversionValue the value of the conversion
+     */
+    public static function runExample(
+        GoogleAdsClient $googleAdsClient,
+        int $customerId,
+        int $conversionActionId,
+        string $gclId,
```
fiboknacky: As discussed above, let's use $gclid, not $gclId

examples/Remarketing/UploadOfflineConversions.php
```diff
@@ -0,0 +1,157 @@
...
+        string $conversionTime,
+        float $conversionValue
+    ) {
+        // Creates a click conversion.
```
fiboknacky: Nit: "Creates a click conversion by specifying currency as USD."

examples/Remarketing/UploadOfflineConversions.php
```diff
@@ -0,0 +1,157 @@
...
+            'conversion_value' => new DoubleValue(['value' => $conversionValue]),
+            'conversion_date_time' => new StringValue(['value' => $conversionTime]),
+            'currency_code' => new StringValue(['value' => 'USD']),
+        ]);
+
+        // Issues a request to upload the click conversion.
+        $conversionUploadServiceClient = $googleAdsClient->getConversionUploadServiceClient();
+        /** @var UploadClickConversionsResponse $response */
+        $response = $conversionUploadServiceClient->uploadClickConversions(
+            $customerId, [$clickConversion], ['partialFailure' => true]
```
fiboknacky: Is this PSR-2 compliant?

examples/Utils/ArgumentNames.php
```diff
@@ -46,6 +49,7 @@ final class ArgumentNames
     const FEED_ITEM_ID = 'feedItemId';
     const FEED_ITEM_IDS = 'feedItemIds';
     const FEED_PLACEHOLDER_FIELD_NAME = 'flightPlaceholderFieldName';
+    const GCL_ID = 'gclId';
```
fiboknacky: GCLID and gclid

examples/Remarketing/UploadOfflineConversions.php
PierrickVoulet: Added the format information

examples/Remarketing/UploadOfflineConversions.php
PierrickVoulet: Good catch, only one argument per line is permitted so I changed

examples/Remarketing/UploadOfflineConversion.php
```diff
@@ -0,0 +1,161 @@
...
+        }
+    }
+
+    /**
+     * Runs the example.
+     *
+     * @param GoogleAdsClient $googleAdsClient the Google Ads API client
+     * @param int $customerId the customer ID
+     * @param int $conversionActionId the ID of the conversion action to upload to
+     * @param string $gclid the GCLID for the conversion (should be newer than 30 days)
```
sundquist: The '(should be newer than 30 days)' implies that it must always be 30 days. However, it's more that the window was set to 30 days on the conversion, so it has to be 30 days here. Could you perhaps add a bit of clarification that it should be newer than the number of days that were set on the conversion window? In this case, the AddConversionAction example had 30 days, so that's why this is 30 days.

PierrickVoulet: Good point!

examples/Remarketing/UploadOfflineConversion.php
```diff
@@ -111,25 +111,26 @@ public static function main()
      * @param GoogleAdsClient $googleAdsClient the Google Ads API client
      * @param int $customerId the customer ID
      * @param int $conversionActionId the ID of the conversion action to upload to
-     * @param string $gclId the GCLID for the conversion (should be newer than 30 days)
+     * @param string $gclid the GCLID for the conversion (should be newer than 30 days)
      * @param string $conversionTime the date and time of the conversion (should be after the
-     * click time)
+     *      click time). The format is "yyyy-mm-dd hh:mm:ss+|-hh:mm", e.g.
```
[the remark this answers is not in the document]
fiboknacky: Reminder for this question before we can merge.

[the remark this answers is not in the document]
PierrickVoulet: I reused the example provided in the PHP doc of the service. My understanding is that "2019-01-01 12:32:45-08:00" follows the "yyyy-mm-dd hh:mm:ss+|-hh:mm" rule:

- "yyyy-mm-dd" => "2019-01-01"
- "hh:mm:ss" => "12:32:45"
- "+|-" => "-"
- "hh:mm" => "08:00"

Issue #237, opened 2019-11-20
Example GetAccountHierarchy lists customer twice when sub-managers are returned


williamjulianvicary:
The code: https://github.com/googleads/google-ads-php/blob/master/examples/AccountManagement/GetAccountHierarchy.php

Is outputting a sub-manager's account within the root-manager account, for example:
Root Manager > Sub Manager > Account

Expected output:
```
Root Manager
-- Sub Manager
---- Account
```

Actual output:
```
Root Manager
-- Account <-- listed twice
-- Sub Manager
---- Account <-- listed twice
```

This is because the sub-manager accounts dive into a level greater than 1, however the query is not constrained.

This can be resolved by setting `WHERE customer_client.level < 2` on the query performed. As this query is recursive and sub-managers are processed separately, we should never require a level greater than 1 in the results from the query.

fiboknacky:
Shouldn't this be covered by [this line](https://github.com/googleads/google-ads-php/blob/master/examples/AccountManagement/GetAccountHierarchy.php#L173)?

Could you please also tell the root manager and account manager IDs for reproduction?

williamjulianvicary:
I'm unable to share the ID's publicly I'm afraid, but it's this [line](https://github.com/googleads/google-ads-php/blob/master/examples/AccountManagement/GetAccountHierarchy.php#L164).

That is causing the trouble, it's adding the account/customer to the `$customerIdsToChildAccounts` map, regardless of whether this account is a sub-account of a manager or not. The line you quote ensures that a _manager_ isn't requested twice, but it does nothing to stop the customers within that manager (level 2+) being added to the output data.

Hope that helps!

fiboknacky:
Is it possible for you to send your IDs to googleadsapi-support@google.com, which is our official support alias, then? It would help a lot if I can reproduce the issue myself. If you're still not comfortable, that's totally fine.
I remember vaguely that we didn't put in `WHERE customer_client.level < 2` because it would restrict the results too much. But I'll check this again.

Thanks for feedback, anyway.

williamjulianvicary:
Perfect - I've pinged the support email with the details requested.

As far as I can tell the `level < 2` should work OK, the root account is extracted from `level = 0`, and sub-customers and sub-managers from `level = 1` - as the code then builds unprocessed accounts where a sub-manager is found with level = 1, once that query runs the level returned for the sub-managers accounts would reset back to 1 (I believe)?

fiboknacky:
[...]

PR #246, opened 2019-12-06
Fixed GetAccountHierarchy so it doesn't print grandchild accounts.

fiboknacky:
Also, add a test to check if the seed customer ID is a test account.
Test accounts are not returned by customer client right now.


fiboknacky:
This is to fix #237

PierrickVoulet (approved):

PR #247, opened 2019-12-09
Added AddSitelinks example

fiboknacky:


examples/Extensions/AddSitelinks.php
```diff
@@ -0,0 +1,339 @@
...
+     * @param int $customerId the client customer ID
+     * @param string $campaignResourceName the resource name of the campaign to target
+     * @return StringValue[] the list of extension feed items' resource names
+     */
+    private static function createExtensionFeedItems(
+        GoogleAdsClient $googleAdsClient,
+        int $customerId,
+        string $campaignResourceName
+    ) {
+        // Creates a site link feed item 1.
```
PierrickVoulet: nit: "**the first** site link feed item" instead of "a site link feed item **1**"?

Same with the other instances

examples/Extensions/AddSitelinks.php
```diff
@@ -0,0 +1,339 @@
+<?php
```
PierrickVoulet: Sometimes you use "sitelink" and "site link". Based on the Google Help terminology, I think it should be "sitelink" everywhere.

Please consider double checking the camel case too if you make changes

examples/Extensions/AddSitelinks.php
```diff
@@ -0,0 +1,339 @@
...
+        if (strtotime('now') > strtotime('20 November')) {
+            $targetYearString = date('Y', strtotime('+1 year'));
+        }
+        $startTimeString = date('Y-m-d H:i:s', mktime(0, 0, 0, 11, 20, intval($targetYearString)));
+        // Use the same year as start time when creating end time.
+        $endTimeString = date('Y-m-d H:i:s', mktime(23, 59, 59, 11, 27, intval($targetYearString)));
+
+        // Targets this sitelink for United States only.
+        // A list of country codes can be referenced here:
+        // https://developers.google.com/adwords/api/docs/appendix/geotargeting
```
PierrickVoulet: Not sure if done on purpose but this link is specific to AdWords API

examples/Extensions/AddSitelinks.php
```diff
@@ -0,0 +1,339 @@
...
+        // Prints some information about the created extension feed items.
+        printf(
+            "Created %d extension feed items with the following resource names:%s",
+            $response->getResults()->count(),
+            PHP_EOL
+        );
+        $createdExtensionFeedItemsResourceNames = [];
+        foreach ($response->getResults() as $addedExtensionFeedItem) {
+            /** @var ExtensionFeedItem $addedExtensionFeedItem */
+            print "{$addedExtensionFeedItem->getResourceName()}" . PHP_EOL;
```
PierrickVoulet: Any reason why you do not use the usual `printf` here?

PierrickVoulet: `$addedExtensionFeedItem->getResourceName()` is called two times, you could use a variable here

examples/Extensions/AddSitelinks.php
```diff
@@ -0,0 +1,339 @@
...
+    private static function createSitelinkFeedItem(string $sitelinkText, string $sitelinkUrl)
+    {
+        return new SitelinkFeedItem([
+            'link_text' => new StringValue(['value' => $sitelinkText]),
+            'final_urls' => [new StringValue(['value' => $sitelinkUrl])],
+        ]);
+    }
+
+    /**
+     * Creates a new AdScheduleInfo with the specified attributes.
```
PierrickVoulet: You use "AdScheduleInfo" (technical name) multiple times in this comment, should not it be "ad schedule info" instead?

examples/Extensions/AddSitelinks.php
```diff
@@ -0,0 +1,339 @@
...
+     *
+     * @param int $day the enum value of day of the week of the AdScheduleInfo
+     * @param int $startHour the starting hour of the AdScheduleInfo
```
PierrickVoulet: Is there any enum we could use like for the minute instead?

PierrickVoulet (changes requested):

examples/Extensions/AddSitelinks.php
fiboknacky: Thanks for catching this. I myself told Devin to use sitelink consistently too.

examples/Extensions/AddSitelinks.php
fiboknacky: We don't have that page in Google Ads API yet.

examples/Extensions/AddSitelinks.php
fiboknacky: Nope. I guess it's too much to create an enum of 24 hours.
That for minute actually has four values for four quarters.

examples/Extensions/AddSitelinks.php
fiboknacky: I was reluctant about this because of the ad schedule info is an uncountable noun.
But just found how to make it more consistent.

examples/Extensions/AddSitelinks.php
fiboknacky: * It's consistent with other examples
* It's not so complicated that we need printf to help (especially after we use local var here instead).

examples/Extensions/AddSitelinks.php
fiboknacky: We don't have this page in Google Ads API yet.

PierrickVoulet (approved):

devchas (approved):

PR #250, opened 2019-12-16
Added support for conversions in errors

PierrickVoulet:
Change-Id: Ie9a2adf3da9a147554583400a4815abcf2b305cb

### DIFF
--- a/tests/Google/Ads/GoogleAds/Util/V2/GoogleAdsErrorsTest.php
+++ b/tests/Google/Ads/GoogleAds/Util/V2/GoogleAdsErrorsTest.php
@@ -38,7 +38,6 @@ class GoogleAdsErrorsTest extends TestCase
     public function setUp()
     {
         $this->failure = $this->createGoogleAdsFailure();
-
         $any = new Any();
         $any->pack($this->failure);
         $this->status = new Status([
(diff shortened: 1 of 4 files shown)


tests/Google/Ads/GoogleAds/Util/V2/GoogleAdsErrorsTest.php
```diff
@@ -38,7 +38,6 @@ class GoogleAdsErrorsTest extends TestCase
     public function setUp()
     {
         $this->failure = $this->createGoogleAdsFailure();
-
```
fiboknacky: Could you revert this change to be consistent with V1?

tests/Google/Ads/GoogleAds/Util/V2/GoogleAdsErrorsTest.php
```diff
@@ -93,14 +92,23 @@ public function testFromFailureWithInvalidOperationIndex()
     public function testFromFailureWithMutateOperation()
     {
         $failureWithMutateOperation = $this->createGoogleAdsFailure("mutate_operations");
-
```
fiboknacky: Ditto

tests/Google/Ads/GoogleAds/Util/V2/GoogleAdsErrorsTest.php
PierrickVoulet: Actually, I removed this line because it was not consistent with V1 (see [here](https://github.com/googleads/google-ads-php/blob/master/tests/Google/Ads/GoogleAds/Util/V1/GoogleAdsErrorsTest.php#L41))

tests/Google/Ads/GoogleAds/Util/V2/GoogleAdsErrorsTest.php
PierrickVoulet: Same (see [here](https://github.com/googleads/google-ads-php/blob/master/tests/Google/Ads/GoogleAds/Util/V1/GoogleAdsErrorsTest.php#L95))

fiboknacky: OK. Thanks for explaining.

PR #255, opened 2020-01-03
Add UploadMediaBundle code example

PierrickVoulet:
Change-Id: I5df08c7f4252ae724114064b82389932348c8361

### DIFF
--- a/examples/Misc/UploadMediaBundle.php
+++ b/examples/Misc/UploadMediaBundle.php
@@ -103,7 +103,6 @@ class UploadMediaBundle
         $mediaFile = new MediaFile([
             'name' => new StringValue(['value' => 'Ad Media Bundle']),
             'type' => MediaType::MEDIA_BUNDLE,
-            'source_url' => new StringValue(['value' => self::BUNDLE_URL]),
             'media_bundle' => new MediaBundle(['data' => $bundleContent])
         ]);
 
--- a/examples/Misc/UploadMediaBundle.php
+++ b/examples/Misc/UploadMediaBundle.php
@@ -1,0 +1,134 @@
+<?php
+/**
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Ads\GoogleAds\Examples\Misc;
+
+require __DIR__ . '/../../vendor/autoload.php';
+
+use GetOpt\GetOpt;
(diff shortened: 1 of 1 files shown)


examples/Misc/UploadMediaBundle.php
```diff
@@ -0,0 +1,134 @@
...
+    public static function runExample(GoogleAdsClient $googleAdsClient, int $customerId)
+    {
+        // Creates an HTML5 zip file media bundle content.
+        $bundleContent = file_get_contents(self::BUNDLE_URL);
+
+        // Creates a media file.
+        $mediaFile = new MediaFile([
+            'name' => new StringValue(['value' => 'Ad Media Bundle']),
+            'type' => MediaType::MEDIA_BUNDLE,
+            'source_url' => new StringValue(['value' => self::BUNDLE_URL]),
```
fiboknacky: Interesting that we need to provide BUNDLE_URL twice here.
source_url is a required field?

examples/Misc/UploadMediaBundle.php
```diff
@@ -0,0 +1,134 @@
...
+            $customerId,
+            [$mediaFileOperation]
+        );
+
+        printf("Added %d media bundles:%s", $response->getResults()->count(), PHP_EOL);
+
+        foreach ($response->getResults() as $addedMediaFile) {
+            /** @var MediaFile $addedMediaFile */
+            printf(
+                "New media bundle added with resource name: '%s'%s",
```
fiboknacky: Nit: Let's use a sentence here "... bundle was added with ..."

PierrickVoulet: I switched to a single message log.

examples/Misc/UploadMediaBundle.php
PierrickVoulet: Good catch, this field is 'Only used for media of type AUDIO and IMAGE' so not needed here.

examples/Misc/UploadMediaBundle.php
```diff
@@ -0,0 +1,128 @@
...
+use Google\Ads\GoogleAds\V2\Resources\MediaFile;
+use Google\Ads\GoogleAds\V2\Services\MediaFileOperation;
+use Google\ApiCore\ApiException;
+use Google\Protobuf\StringValue;
+
+/** This example uploads an HTML5 zip file as a media bundle. */
+class UploadMediaBundle
+{
+    const CUSTOMER_ID = 'INSERT_CUSTOMER_ID_HERE';
+    const BUNDLE_URL = 'https://goo.gl/9Y7qI2';
```
devchas: Is there any reason this should be defined at the class level instead of in the method?

PierrickVoulet: It is used only once in the code so it could be used directly in the method as you suggest.

I believe it is better to have it at class level though because it is a constant, it is more readable and it is consistent with similar code examples like [UploadImageAsset](https://github.com/googleads/google-ads-php/pull/254) and [UploadImage](https://github.com/googleads/google-ads-php/blob/master/examples/Misc/UploadImage.php).

WDYT?

fiboknacky: As it's not supposed to change frequently, I think it's fine to be constant here.
If we want to be more accurate, we could make it private const (probably supported since PHP 7?)
But that it's not a big deal to me.

PierrickVoulet: @devchas if no further concerns regarding this on your end, I will consider this comment closed and will open an issue to change for private constants across all examples.

PR #257, opened 2020-01-08
Update AddDynamicFeed advanced operation example

oniltos:
When running the example the feed and feed items were created, but the Page URL and the Label contents were empty. The reason behind the empty values lies in the use of "string_value" property in the constructor of FeedItemAttributeValue class. Since the feed attributes of created feed have URL_LIST and STRING_LIST as their type, the Page URL and the Label must be arrays. 
This commit changes the key from "string_value" to "string_values" in the calls of FeedItemAttributeValue class on createFeedItems() method and put the property values inside arrays.


oniltos:
@googlebot I signed it!

PR #262, opened 2020-01-22
Add the ValidateTextAd code example

PierrickVoulet:
Change-Id: Ie90a4f0668cf82d6dd7d5ffa369d78d87d8ff6ef

### DIFF
--- a/examples/CampaignManagement/ValidateTextAd.php
+++ b/examples/CampaignManagement/ValidateTextAd.php
@@ -1,0 +1,197 @@
+<?php
+/**
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Ads\GoogleAds\Examples\CampaignManagement;
+
+require __DIR__ . '/../../vendor/autoload.php';
+
+use GetOpt\GetOpt;
+use Google\Ads\GoogleAds\Examples\Utils\ArgumentNames;
+use Google\Ads\GoogleAds\Examples\Utils\ArgumentParser;
+use Google\Ads\GoogleAds\Lib\OAuth2TokenBuilder;
+use Google\Ads\GoogleAds\Lib\V2\GoogleAdsClient;
+use Google\Ads\GoogleAds\Lib\V2\GoogleAdsClientBuilder;
+use Google\Ads\GoogleAds\Lib\V2\GoogleAdsException;
+use Google\Ads\GoogleAds\Util\V2\ResourceNames;
+use Google\Ads\GoogleAds\V2\Common\ExpandedTextAdInfo;
+use Google\Ads\GoogleAds\V2\Common\PolicyTopicEntry;
+use Google\Ads\GoogleAds\V2\Enums\AdGroupAdStatusEnum\AdGroupAdStatus;
+use Google\Ads\GoogleAds\V2\Enums\PolicyTopicEntryTypeEnum\PolicyTopicEntryType;
+use Google\Ads\GoogleAds\V2\Errors\GoogleAdsError;
+use Google\Ads\GoogleAds\V2\Errors\PolicyFindingErrorEnum\PolicyFindingError;
+use Google\Ads\GoogleAds\V2\Errors\PolicyViolationErrorEnum\PolicyViolationError;
+use Google\Ads\GoogleAds\V2\Resources\Ad;
+use Google\Ads\GoogleAds\V2\Resources\AdGroupAd;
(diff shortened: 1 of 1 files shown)


examples/CampaignManagement/ValidateTextAd.php
```diff
@@ -0,0 +1,197 @@
...
+            printf("Expanded text ad validated successfully.%s", PHP_EOL);
+        } catch (GoogleAdsException $googleAdsException) {
+            // This block will be hit if there is a validation error from the server.
+            printf("There were validation error(s) while adding expanded text ad.%s", PHP_EOL);
+
+            $count = 1;
+            foreach ($googleAdsException->getGoogleAdsFailure()->getErrors() as $googleAdsError) {
+                /** @var GoogleAdsError $googleAdsError */
+                if ($googleAdsError->getErrorCode()->getPolicyFindingError()
+                    == PolicyFindingError::POLICY_FINDING
```
fiboknacky: Use ==
And usually we put == on the previous line, but || and && in the next line.

examples/CampaignManagement/ValidateTextAd.php
```diff
@@ -0,0 +1,197 @@
...
+                            );
+                        }
+                    } else if ($googleAdsError->getDetails()->getPolicyViolationDetails()) {
+                        $details = $googleAdsError->getDetails()->getPolicyViolationDetails();
+                        printf(
+                            "%d) Policy violation with name '%s' and isExemptable %s" .
+                            " was found.%s",
+                            $count,
+                            $details->getExternalPolicyName(),
+                            var_export($details->getIsExemptible(), true),
```
fiboknacky: Can you explain why we need var_export for this? To return True and False?

examples/CampaignManagement/ValidateTextAd.php
```diff
@@ -0,0 +1,197 @@
...
+                $customerId,
+                [$adGroupAdOperation],
+                ['validateOnly' => true]
+            );
+
+            // Since validation is ON, result will be null.
+            printf("Expanded text ad validated successfully.%s", PHP_EOL);
+        } catch (GoogleAdsException $googleAdsException) {
+            // This block will be hit if there is a validation error from the server.
+            printf("There were validation error(s) while adding expanded text ad.%s", PHP_EOL);
```
fiboknacky: Just double check. This fits in 100 chars, right?

examples/CampaignManagement/ValidateTextAd.php
PierrickVoulet: Yes, do you know an elegant way of printing boolean value with printf?

examples/CampaignManagement/ValidateTextAd.php
PierrickVoulet: Yes, it does.

examples/CampaignManagement/ValidateTextAd.php
fiboknacky: I think there is no one way to do that.
Previously, we usually use [this way](https://stackoverflow.com/a/2795180) though, which is personally a bit better because the code readers wouldn't need to know about var_export (and its "true" parameter).

examples/CampaignManagement/ValidateTextAd.php
PierrickVoulet: I updated, is it right?

fiboknacky: That's right.

examples/CampaignManagement/ValidateTextAd.php
PierrickVoulet: SGTM, updated

examples/CampaignManagement/ValidateTextAd.php
```diff
@@ -0,0 +1,197 @@
...
+            $count = 1;
+            foreach ($googleAdsException->getGoogleAdsFailure()->getErrors() as $googleAdsError) {
+                /** @var GoogleAdsError $googleAdsError */
+                if ($googleAdsError->getErrorCode()->getPolicyFindingError() ==
+                    PolicyFindingError::POLICY_FINDING
+                    || $googleAdsError->getErrorCode()->getPolicyViolationError() ==
+                    PolicyViolationError::POLICY_ERROR) {
+                    // Only one of PolicyFindingDetails or PolicyViolationDetails will be
+                    // populated. PolicyViolationDetails is used by some ad formats, and
+                    // PolicyFindingDetails by others.
```
fiboknacky: In this example, I think we should scope down to only one class either PolicyViolationDetails or PolicyFindingDetails. You can just link to my [created guide for details](https://developers.google.com/google-ads/api/docs/policy-exemption/overview?hl=en).
As ETA only returns errors in PolicyFindingDetails, we should remove the part related to PolicyViolationErrors.

PierrickVoulet: Makes sense to me, I replicated what is in the [golden](https://github.com/googleads/google-ads-dotnet/blob/master/examples/CampaignManagement/ValidateTextAd.cs).

@AnashOommen : WDYT? Any chance to update the Golden soon (before other client libraries start porting)?

AnashOommen: I'll fix the golden and point to https://developers.google.com/google-ads/api/docs/policy-exemption/overview for more details.

AnashOommen: https://github.com/googleads/google-ads-dotnet/pull/148

PierrickVoulet: Thanks @AnashOommen , I updated too

PR #274, opened 2020-02-18
Added AddListingScope example.

fiboknacky:


examples/ShoppingAds/AddListingScope.php
```diff
@@ -0,0 +1,184 @@
...
+    public static function runExample(
+        GoogleAdsClient $googleAdsClient,
+        int $customerId,
+        int $campaignId
+    ) {
+
+        // A listing scope allows you to filter the products that will be included in a given
+        // campaign. You can specify multiple dimensions with conditions that must be met for a
+        // product to be included in a campaign.
+        // A typical ListingScope might only have a few dimensions. This example demonstrates a
```
PierrickVoulet: nit: "ListingScope" is a type, do we want to change for english words like "listing scope" to make it consistent with all code example comments?

examples/ShoppingAds/AddListingScope.php
```diff
@@ -0,0 +1,184 @@
...
+        // range of different dimensions you could use.
+        $listingScopeInfo = new ListingScopeInfo([
+            'dimensions' => [
+                // Creates a listing brand info set to "google".
+                new ListingDimensionInfo([
+                    'listing_brand'
+                    => new ListingBrandInfo(['value' => new StringValue(['value' => 'google'])])
```
PierrickVoulet: In the next major release, ListingBrandInfo will be renamed to ProductBrandInfo. Like the AddShoppingProductListingGroupTree code example, this will need to be fixed.

What about waiting for the next release before reworking/merging this code example?

examples/ShoppingAds/AddListingScope.php
```diff
@@ -0,0 +1,184 @@
...
+            'campaign' =>
+                new StringValue(['value' => ResourceNames::forCampaign($customerId, $campaignId)]),
+            'listing_scope' => $listingScopeInfo
+        ]);
+
+        // Issues a mutate request to create a campaign criterion on the server and print its info.
+        $campaignCriterionServiceClient = $googleAdsClient->getCampaignCriterionServiceClient();
+        $response = $campaignCriterionServiceClient->mutateCampaignCriteria(
+            $customerId,
+            [new CampaignCriterionOperation(['create' => $campaignCriterion])]
```
PierrickVoulet: In our other code examples, the operation is created and stored in a dedicated variable. You might want to do the same here for consistency.

PierrickVoulet (changes requested):

examples/ShoppingAds/AddListingScope.php
fiboknacky: Actually, this is more consistent with all other places, i.e., when creating an object, we now specify a field directly (like creating ListingDimensionInfo above).
We don't use `setProductType()` method, for instance.

Because of the above reason and its conciseness, I actually prefer this style.
But I just don't feel necessary to change all the examples to follow this.

I'll change this to match other examples for now, but please also let me know what you think in the long run. :)

PierrickVoulet: I prefer this one too: easily understandable, readable and more concise.

PierrickVoulet (approved):

jradcliff (approved):

PR #275, opened 2020-02-18
Add AddAppCampaign code example

PierrickVoulet:
Change-Id: I74a6dac7e547e53a596d604fc0f0e057567f33fa

### DIFF
--- a/examples/AdvancedOperations/AddAppCampaign.php
+++ b/examples/AdvancedOperations/AddAppCampaign.php
@@ -403,12 +403,20 @@ class AddAppCampaign
                 // text_ad, gmail_ad, etc.
                 'app_ad' => new AppAdInfo([
                     'headlines' => [
-                        self::createAdTextAsset('A cool puzzle game'),
-                        self::createAdTextAsset('Remove connected blocks')
+                        new AdTextAsset([
+                            'text' => new StringValue(['value' => 'A cool puzzle game'])
+                        ]),
+                        new AdTextAsset([
+                            'text' => new StringValue(['value' => 'Remove connected blocks'])
+                        ])
                     ],
                     'descriptions' => [
-                        self::createAdTextAsset('3 difficulty levels'),
-                        self::createAdTextAsset('4 colorful fun skins')
+                        new AdTextAsset([
+                            'text' => new StringValue(['value' => '3 difficulty levels'])
+                        ]),
+                        new AdTextAsset([
+                            'text' => new StringValue(['value' => '4 colorful fun skins'])
+                        ])
                     ]
                     // Optional: You can set up to 20 image assets for your campaign.
                     // 'images' => [INSERT_AD_IMAGE_ASSET_ID(s)_HERE]
@@ -430,20 +438,6 @@ class AddAppCampaign
             $createdAdGroupAdResourceName,
             PHP_EOL
         );
-    }
-
-    /**
-     * Creates an ad text asset for a given text.
-     *
-     * @param string $text the text to create the ad asset for
-     * @return string the created ad text asset
-     */
-    private static function createAdTextAsset(string $text)
-    {
-        // Creates and returns an ad text asset.
-        return new AdTextAsset([
-            'text' => new StringValue(['value' => $text])
-        ]);
     }
 }
 
--- a/examples/AdvancedOperations/AddAppCampaign.php
+++ b/examples/AdvancedOperations/AddAppCampaign.php
@@ -27,7 +27,6 @@ use Google\Ads\GoogleAds\Lib\V2\GoogleAdsClientBuilder;
 use Google\Ads\GoogleAds\Lib\V2\GoogleAdsException;
 use Google\Ads\GoogleAds\Lib\OAuth2TokenBuilder;
 use Google\Ads\GoogleAds\Util\V2\ResourceNames;
-use Google\Ads\GoogleAds\V1\Enums\BiddingStrategyTypeEnum\BiddingStrategyType;
 use Google\Ads\GoogleAds\V2\Common\AdTextAsset;
 use Google\Ads\GoogleAds\V2\Common\AppAdInfo;
 use Google\Ads\GoogleAds\V2\Common\LanguageInfo;
@@ -67,7 +66,7 @@ use Google\Protobuf\StringValue;
  * https://developers.google.com/google-ads/api/docs/app-campaigns/overview
  *
  * To get campaigns, run the GetCampaigns.php example.
- * To upload image assets for this campaign, run the UploadImage.php example.
+ * To upload image assets for this campaign, run the UploadImageAsset.php example.
  */
 class AddAppCampaign
 {
@@ -131,7 +130,7 @@ class AddAppCampaign
         GoogleAdsClient $googleAdsClient,
         int $customerId
     ) {
-        // Creates the budget for the campaign.
+        // Creates a budget for the campaign.
         $budgetResourceName = self::createBudget($googleAdsClient, $customerId);
 
         // Creates the campaign.
@@ -144,7 +143,7 @@ class AddAppCampaign
         // Sets campaign targeting.
         self::setCampaignTargetingCriteria($googleAdsClient, $customerId, $campaignResourceName);
 
-        // Creates Ad Group and App Ad.
+        // Creates an ad group and an App ad.
         self::createAppAd($googleAdsClient, $customerId, $campaignResourceName);
     }
 
@@ -182,7 +181,7 @@ class AddAppCampaign
 
         $createdCampaignBudgetResourceName = $response->getResults()[0]->getResourceName();
         printf(
-            "Created campaign budget with resource name: '%s'%s",
+            "Created campaign budget with resource name: '%s'.%s",
             $createdCampaignBudgetResourceName,
             PHP_EOL
         );
@@ -217,21 +216,17 @@ class AddAppCampaign
             // eligible to appear on multiple channels.
             'advertising_channel_type' => AdvertisingChannelType::MULTI_CHANNEL,
             'advertising_channel_sub_type' => AdvertisingChannelSubType::APP_CAMPAIGN,
-            // Define the bidding strategy during campaign creation. An App campaign
-            // cannot use a portfolio bidding strategy.
-            // App campaigns only support the TARGET_CPA bidding strategy.
-            'bidding_strategy_type' => BiddingStrategyType::TARGET_CPA,
-            // Set the target CPA to $1 / app install.
+            // Sets the target CPA to $1 / app install.
             //
             // campaign_bidding_strategy is a 'oneof' message so setting target_cpa
             // is mutually exclusive with other bidding strategies such as
             // manual_cpc, commission, maximize_conversions, etc.
             // See https://developers.google.com/google-ads/api/reference/rpc
-            // under current version / resources / Campaign
+            // under current version / resources / Campaign.
             'target_cpa' => new TargetCpa([
                 'target_cpa_micros' => new Int64Value(['value' => 1000000])
             ]),
-            // Set the App Campaign Settings.
+            // Sets the App campaign settings.
             'app_campaign_setting' => new AppCampaignSetting([
                 'app_id' => new StringValue(['value' => 'com.google.android.apps.adwords']),
                 'app_store' => AppCampaignAppStore::GOOGLE_APP_STORE,
@@ -239,7 +234,7 @@ class AddAppCampaign
                 'bidding_strategy_goal_type' =>
                     AppCampaignBiddingStrategyGoalType::OPTIMIZE_INSTALLS_TARGET_INSTALL_COST
             ]),
-            // Optional fields
+            // Optional fields.
             'start_date' => new StringValue(['value' => date('Ymd', strtotime('+1 day'))]),
             'end_date' => new StringValue(['value' => date('Ymd', strtotime('+365 days'))])
             // If you select the
@@ -263,7 +258,11 @@ class AddAppCampaign
         $response = $campaignServiceClient->mutateCampaigns($customerId, [$campaignOperation]);
 
         $createdCampaignResourceName = $response->getResults()[0]->getResourceName();
-        printf("Created App campaign '%s'%s", $createdCampaignResourceName, PHP_EOL);
+        printf(
+            "Created App campaign with resource name: '%s'.%s",
+            $createdCampaignResourceName,
+            PHP_EOL
+        );
 
         return $createdCampaignResourceName;
     }
@@ -323,8 +322,8 @@ class AddAppCampaign
                 'campaign' => new StringValue(['value' => $campaignResourceName]),
                 'type' => CriterionType::LANGUAGE,
                 'language' => new LanguageInfo([
-                    'language_constant' => new StringValue(['value' =>
-                        ResourceNames::forLanguageConstant($languageId)
+                    'language_constant' => new StringValue([
+                        'value' => ResourceNames::forLanguageConstant($languageId)
                     ])
                 ])
             ]);
@@ -336,23 +335,32 @@ class AddAppCampaign
             $campaignCriterionOperations[] = $campaignCriterionOperation;
         }
 
-        // Submits the criteria operations.
+        // Submits the criteria operations and prints their information.
         $campaignCriterionServiceClient = $googleAdsClient->getCampaignCriterionServiceClient();
         $response = $campaignCriterionServiceClient->mutateCampaignCriteria(
             $customerId,
             $campaignCriterionOperations
         );
 
-        $createdCampaignCriteriaResourceName = $response->getResults()[0]->getResourceName();
-        printf("Created Campaign Criteria '%s'%s", $createdCampaignCriteriaResourceName, PHP_EOL);
+        printf(
+            "Created %d Campaign Criteria with resource names:%s",
+            $response->getResults()->count(),
+            PHP_EOL
+        );
+
+        foreach ($response->getResults() as $createdCampaignCriterion) {
+            /** @var CampaignCriterion $createdCampaignCriterion */
+            printf("\t%s%s", $createdCampaignCriterion->getResourceName(), PHP_EOL);
+        }
     }
 
     /**
-     * Creates an ad group and associated app ap for a given campaign.
+     * Creates an ad group and associated App ad for a given campaign.
      *
      * @param GoogleAdsClient $googleAdsClient the Google Ads API client
      * @param int $customerId the customer ID
-     * @param string $campaignResourceName the resource name of the campaign to be modified
+     * @param string $campaignResourceName the resource name of the campaign to add ad group and
+     *      App ad to
      */
     private static function createAppAd(
         GoogleAdsClient $googleAdsClient,
@@ -362,8 +370,8 @@ class AddAppCampaign
         // Creates an ad group.
         // Note that the ad group type must not be set.
         // Since the advertising_channel_sub_type is APP_CAMPAIGN,
-        //   1- you cannot override bid settings at the ad group level.
-        //   2- you cannot add ad group criteria
+        //   1. you cannot override bid settings at the ad group level.
+        //   2. you cannot add ad group criteria.
         $adGroup = new AdGroup([
             'name' => new StringValue(['value' => 'Earth to Mars cruises ' . uniqid()]),
             'status' => AdGroupStatus::ENABLED,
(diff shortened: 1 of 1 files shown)


examples/AdvancedOperations/AddAppCampaign.php
```diff
@@ -0,0 +1,443 @@
...
+        $campaignResourceName = self::createCampaign(
+            $googleAdsClient,
+            $customerId,
+            $budgetResourceName
+        );
+
+        // Sets campaign targeting.
+        self::setCampaignTargetingCriteria($googleAdsClient, $customerId, $campaignResourceName);
+
+        // Creates Ad Group and App Ad.
```
fiboknacky: Should be "Creates an ad group and an app ad."

examples/AdvancedOperations/AddAppCampaign.php
```diff
@@ -0,0 +1,443 @@
...
+     * Runs the example.
+     *
+     * @param GoogleAdsClient $googleAdsClient the Google Ads API client
+     * @param int $customerId the customer ID
+     */
+    public static function runExample(
+        GoogleAdsClient $googleAdsClient,
+        int $customerId
+    ) {
+        // Creates the budget for the campaign.
```
fiboknacky: Nit: I'm not a grammar expert, but should we use "a" instead of "the" here?
In the PHPDoc of this method, we say "Creates a budget".

examples/AdvancedOperations/AddAppCampaign.php
```diff
@@ -0,0 +1,443 @@
...
+            'status' => CampaignStatus::PAUSED,
+            // All App campaigns have an advertising_channel_type of
+            // MULTI_CHANNEL to reflect the fact that ads from these campaigns are
+            // eligible to appear on multiple channels.
+            'advertising_channel_type' => AdvertisingChannelType::MULTI_CHANNEL,
+            'advertising_channel_sub_type' => AdvertisingChannelSubType::APP_CAMPAIGN,
+            // Define the bidding strategy during campaign creation. An App campaign
+            // cannot use a portfolio bidding strategy.
+            // App campaigns only support the TARGET_CPA bidding strategy.
+            'bidding_strategy_type' => BiddingStrategyType::TARGET_CPA,
```
fiboknacky: This is a read-only field. Please remove.

examples/AdvancedOperations/AddAppCampaign.php
```diff
@@ -0,0 +1,443 @@
...
+            // Set the target CPA to $1 / app install.
+            //
+            // campaign_bidding_strategy is a 'oneof' message so setting target_cpa
+            // is mutually exclusive with other bidding strategies such as
+            // manual_cpc, commission, maximize_conversions, etc.
+            // See https://developers.google.com/google-ads/api/reference/rpc
+            // under current version / resources / Campaign
```
fiboknacky: Nit: End with a period.

examples/AdvancedOperations/AddAppCampaign.php
```diff
@@ -0,0 +1,443 @@
...
+            'target_cpa' => new TargetCpa([
+                'target_cpa_micros' => new Int64Value(['value' => 1000000])
+            ]),
+            // Set the App Campaign Settings.
```
fiboknacky: Nit: campaign is not needed to be capital letter.

examples/AdvancedOperations/AddAppCampaign.php
```diff
@@ -0,0 +1,443 @@
...
+            'app_campaign_setting' => new AppCampaignSetting([
+                'app_id' => new StringValue(['value' => 'com.google.android.apps.adwords']),
+                'app_store' => AppCampaignAppStore::GOOGLE_APP_STORE,
+                // Optional: Optimize this campaign for getting new users for your app.
+                'bidding_strategy_goal_type' =>
+                    AppCampaignBiddingStrategyGoalType::OPTIMIZE_INSTALLS_TARGET_INSTALL_COST
+            ]),
+            // Optional fields
```
fiboknacky: End with a period.

examples/AdvancedOperations/AddAppCampaign.php
```diff
@@ -0,0 +1,443 @@
...
+        // Creates a campaign operation.
+        $campaignOperation = new CampaignOperation();
+        $campaignOperation->setCreate($campaign);
+
+        // Submits the campaign operation and prints the results.
+        $campaignServiceClient = $googleAdsClient->getCampaignServiceClient();
+        $response = $campaignServiceClient->mutateCampaigns($customerId, [$campaignOperation]);
+
+        $createdCampaignResourceName = $response->getResults()[0]->getResourceName();
+        printf("Created App campaign '%s'%s", $createdCampaignResourceName, PHP_EOL);
```
fiboknacky: Need a period before last %s

examples/AdvancedOperations/AddAppCampaign.php
```diff
@@ -0,0 +1,443 @@
...
+        // Submits the campaign budget operation to add the campaign budget.
+        $campaignBudgetServiceClient = $googleAdsClient->getCampaignBudgetServiceClient();
+        $response = $campaignBudgetServiceClient->mutateCampaignBudgets(
+            $customerId,
+            [$campaignBudgetOperation]
+        );
+
+        $createdCampaignBudgetResourceName = $response->getResults()[0]->getResourceName();
+        printf(
+            "Created campaign budget with resource name: '%s'%s",
```
fiboknacky: Need a period before last %s

examples/AdvancedOperations/AddAppCampaign.php
```diff
@@ -0,0 +1,443 @@
...
+            // Creates a campaign criterion operation.
+            $campaignCriterionOperation = new CampaignCriterionOperation();
+            $campaignCriterionOperation->setCreate($campaignCriterion);
+
+            $campaignCriterionOperations[] = $campaignCriterionOperation;
+        }
+
+        // Creates the language campaign criteria.
+        $languageIds = [
+            1000, // English
```
fiboknacky: Probably it's fine, but could you double check if this style complies with PSR-2?

examples/AdvancedOperations/AddAppCampaign.php
```diff
@@ -0,0 +1,443 @@
...
+            1003  // Spanish
+        ];
+        foreach ($languageIds as $languageId) {
+            // Creates a campaign criterion.
+            $campaignCriterion = new CampaignCriterion([
+                'campaign' => new StringValue(['value' => $campaignResourceName]),
+                'type' => CriterionType::LANGUAGE,
+                'language' => new LanguageInfo([
+                    'language_constant' => new StringValue(['value' =>
```
fiboknacky: This is probably against PSR-2.

examples/AdvancedOperations/AddAppCampaign.php
```diff
@@ -0,0 +1,443 @@
...
+            ]);
+
+            // Creates a campaign criterion operation.
+            $campaignCriterionOperation = new CampaignCriterionOperation();
+            $campaignCriterionOperation->setCreate($campaignCriterion);
+
+            $campaignCriterionOperations[] = $campaignCriterionOperation;
+        }
+
+        // Submits the criteria operations.
```
fiboknacky: Let's make it consistent with other functions in this example, e.g., add a bit more comments like "and print their information."

We probably don't need to make it consistent across files. But at least for this one, let's make all comments as consistent as possible.

examples/AdvancedOperations/AddAppCampaign.php
```diff
@@ -0,0 +1,443 @@
...
+        $campaignCriterionServiceClient = $googleAdsClient->getCampaignCriterionServiceClient();
+        $response = $campaignCriterionServiceClient->mutateCampaignCriteria(
+            $customerId,
+            $campaignCriterionOperations
+        );
+
+        $createdCampaignCriteriaResourceName = $response->getResults()[0]->getResourceName();
```
fiboknacky: This is probably wrong. We created many criteria. Why do we print only one?
I think Fan and David also fixed this. Could you see the final version of the file?

https://github.com/googleads/google-ads-python/pull/220/files

examples/AdvancedOperations/AddAppCampaign.php
```diff
@@ -0,0 +1,443 @@
...
+        printf("Created Campaign Criteria '%s'%s", $createdCampaignCriteriaResourceName, PHP_EOL);
+    }
+
+    /**
+     * Creates an ad group and associated app ap for a given campaign.
```
fiboknacky: associated app ap -> associated App ad.

examples/AdvancedOperations/AddAppCampaign.php
```diff
@@ -0,0 +1,443 @@
...
+     *
+     * @param GoogleAdsClient $googleAdsClient the Google Ads API client
+     * @param int $customerId the customer ID
+     * @param string $campaignResourceName the resource name of the campaign to be modified
```
fiboknacky: Last part should be "... of the campaign to add ad group and App ad to"

examples/AdvancedOperations/AddAppCampaign.php
```diff
@@ -0,0 +1,443 @@
...
+     */
+    private static function createAppAd(
+        GoogleAdsClient $googleAdsClient,
+        int $customerId,
+        string $campaignResourceName
+    ) {
+        // Creates an ad group.
+        // Note that the ad group type must not be set.
+        // Since the advertising_channel_sub_type is APP_CAMPAIGN,
+        //   1- you cannot override bid settings at the ad group level.
```
fiboknacky: Really a nit: I would just use "1." and "2."

examples/AdvancedOperations/AddAppCampaign.php
```diff
@@ -0,0 +1,443 @@
...
+            'name' => new StringValue(['value' => 'Earth to Mars cruises ' . uniqid()]),
+            'status' => AdGroupStatus::ENABLED,
+            'campaign' => new StringValue(['value' => $campaignResourceName])
+        ]);
+
+        // Creates an ad group operation.
+        $adGroupOperation = new AdGroupOperation();
+        $adGroupOperation->setCreate($adGroup);
+
+        // Submits the ad group operation to add the ad group.
```
fiboknacky: Ditto about comments.

examples/AdvancedOperations/AddAppCampaign.php
```diff
@@ -0,0 +1,443 @@
...
+        $adGroupServiceClient = $googleAdsClient->getAdGroupServiceClient();
+        $response = $adGroupServiceClient->mutateAdGroups($customerId, [$adGroupOperation]);
+
+        $createdAdGroupResourceName = $response->getResults()[0]->getResourceName();
+        printf(
+            "Ad Group created '%s'%s",
```
fiboknacky: "Created an ad group with resource name. '%s'.%s"

examples/AdvancedOperations/AddAppCampaign.php
```diff
@@ -0,0 +1,443 @@
...
+                    // 'images' => [INSERT_AD_IMAGE_ASSET_ID(s)_HERE]
+                ])
+            ])
+        ]);
+
+        // Creates an ad group ad operation.
+        $adGroupAdOperation = new AdGroupAdOperation();
+        $adGroupAdOperation->setCreate($adGroupAd);
+
+        // Submits the ad group ad operation to add the ad group ad.
```
fiboknacky: Ditto about comments.

examples/AdvancedOperations/AddAppCampaign.php
```diff
@@ -0,0 +1,443 @@
...
+        $adGroupAdServiceClient = $googleAdsClient->getAdGroupAdServiceClient();
+        $response = $adGroupAdServiceClient->mutateAdGroupAds($customerId, [$adGroupAdOperation]);
+
+        $createdAdGroupAdResourceName = $response->getResults()[0]->getResourceName();
+        printf(
+            "Ad Group App Ad created '%s'%s",
```
fiboknacky: "Created an ad group ad with resource name '%s'.%s"

OR

"Created an app ad with resource name '%s'.%s"

examples/AdvancedOperations/AddAppCampaign.php
```diff
@@ -0,0 +1,443 @@
...
+    }
+
+    /**
+     * Creates an ad text asset for a given text.
+     *
+     * @param string $text the text to create the ad asset for
+     * @return string the created ad text asset
+     */
+    private static function createAdTextAsset(
+        string $text
```
fiboknacky: This can fit one line.

examples/AdvancedOperations/AddAppCampaign.php
```diff
@@ -0,0 +1,443 @@
...
+use Google\Protobuf\StringValue;
+
+/**
+ * This example adds an App campaign.
+ *
+ * For guidance regarding App Campaigns, see:
+ * https://developers.google.com/google-ads/api/docs/app-campaigns/overview
+ *
+ * To get campaigns, run the GetCampaigns.php example.
+ * To upload image assets for this campaign, run the UploadImage.php example.
```
fiboknacky: UploadImageAssets.

examples/AdvancedOperations/AddAppCampaign.php
PierrickVoulet: PHPCS runs without errors/warnings

examples/AdvancedOperations/AddAppCampaign.php
PierrickVoulet: It passed PHPCS without errors/warnings but I changed for more readable style.

examples/AdvancedOperations/AddAppCampaign.php
PierrickVoulet: The current version of the golden [does it like this](https://github.com/googleads/google-ads-python/blob/master/examples/advanced_operations/add_app_campaign.py#L227) so I changed for a listing instead.

examples/AdvancedOperations/AddAppCampaign.php
AnashOommen: App Ad.

Also, scan to make sure you are using the same casing everywhere. E.g. I see both **c**ampaign and **C**ampaign in the same code example. Make sure you use a consistent style.

examples/AdvancedOperations/AddAppCampaign.php
```diff
@@ -0,0 +1,443 @@
...
+
+        $createdAdGroupResourceName = $response->getResults()[0]->getResourceName();
+        printf(
+            "Ad Group created '%s'%s",
+            $createdAdGroupResourceName,
+            PHP_EOL
+        );
+
+        // Creates an ad group ad.
+        $adGroupAd = new AdGroupAd([
```
AnashOommen: Could you split the method into two -- one for adgroup creation and the other for ad creation? The method here is named createAppAd, but it creates an adgroup and an ad, which doesn't sound right.

examples/AdvancedOperations/AddAppCampaign.php
```diff
@@ -0,0 +1,443 @@
...
+        );
+    }
+
+    /**
+     * Creates an ad text asset for a given text.
+     *
+     * @param string $text the text to create the ad asset for
+     * @return string the created ad text asset
+     */
+    private static function createAdTextAsset(
```
AnashOommen: This is a small function, you can probably get rid of it and inline the code instead.

examples/AdvancedOperations/AddAppCampaign.php
fiboknacky: ad should be small letter. Only "App" should be in a capital letter.

fiboknacky: ad should be small letter. Only "App" should be in a capital letter.

examples/AdvancedOperations/AddAppCampaign.php
PierrickVoulet: I agree. @fiboknacky this sounds like a major change to me, I am happy to do it with PHP but not sure that it is the best process to follow: I think the golden needs to be updated first?

PierrickVoulet: I agree. @fiboknacky this sounds like a major change to me, I am happy to do it with PHP but not sure that it is the best process to follow: I think the golden needs to be updated first?

examples/AdvancedOperations/AddAppCampaign.php
PierrickVoulet: I agree. @fiboknacky this sounds like a major change to me, I am happy to do it with PHP but not sure it is the best process to follow: I thin the golden needs to be updated first?

fiboknacky: For this one, I think it's a small detail so you can just go ahead and change only in this library.
It'll not be trivial if the method is really used or planned to be used in a guide.

examples/AdvancedOperations/AddAppCampaign.php
fiboknacky: For this one, the reason why I didn't speak up (actually I noticed it too) was I thought that probably the golden CL author would want this to match the guide, which is also divided into two parts: 1) creating a campaign, 2) creating an ad group and ad.

But I see no harm to do what Anash suggested too.
For this one, we would need to reach out to the golden CL author (and other libs that already implemented it) to change too.

PierrickVoulet: @wihl WDYT?

wihl: Yes @fiboknacky is correct. This was done explicitly to mirror the guide (as were many of the comments). As the guides reference the examples, the examples should IMO be consistent with the guides.

wihl: Yes @fiboknacky is correct. This was done explicitly to mirror the guide (as were many of the comments). As the guides reference the examples, the examples should IMO be consistent with the guides.

wihl: Yes @fiboknacky is correct. This was done explicitly to mirror the guide (as were many of the comments). As the guides reference the examples, the examples should IMO be consistent with the guides.

PierrickVoulet: SGTM. @AnashOommen you OK with no split?

PierrickVoulet: @AnashOommen I did the change as discussed offline.
@fiboknacky I will notify the other client lib owners of the change as soon as it is reviewed and merged.

examples/AdvancedOperations/AddAppCampaign.php
```diff
@@ -0,0 +1,466 @@
...
+        string $campaignResourceName
+    ) {
+        // Creates an ad group.
+        // Note that the ad group type must not be set.
+        // Since the advertising_channel_sub_type is APP_CAMPAIGN,
+        //   1. you cannot override bid settings at the ad group level.
+        //   2. you cannot add ad group criteria.
+        $adGroup = new AdGroup([
+            'name' => new StringValue(['value' => 'Earth to Mars cruises ' . uniqid()]),
+            'status' => AdGroupStatus::ENABLED,
```
devchas: Should this be PAUSED to conform to other examples? We usually set campaign and ad group status to PAUSED as a best practice.

examples/AdvancedOperations/AddAppCampaign.php
```diff
@@ -0,0 +1,466 @@
...
+                    'descriptions' => [
+                        new AdTextAsset([
+                            'text' => new StringValue(['value' => '3 difficulty levels'])
+                        ]),
+                        new AdTextAsset([
+                            'text' => new StringValue(['value' => '4 colorful fun skins'])
+                        ])
+                    ]
+                    // Optional: You can set up to 20 image assets for your campaign.
+                    // 'images' => [INSERT_AD_IMAGE_ASSET_ID(s)_HERE]
```
devchas: Is this correct? I believe the images field takes an image object that accepts the asset ID. I don't think you can just pass an array of IDs.

examples/AdvancedOperations/AddAppCampaign.php
wihl: Note that the entire campaign is paused. This is based on the previous AdWords example that had a paused campaign and an enabled Ad Group.

examples/AdvancedOperations/AddAppCampaign.php
wihl: Good catch. This should be [INSERT_AD_IMAGE_RESOURCE_NAME(s)_HERE](https://github.com/googleads/google-ads-python/blob/667ac6d79c6c70480e47a78b5feb045c968d99cb/examples/advanced_operations/add_app_campaign.py#L275)

PierrickVoulet: Fix now under review https://github.com/googleads/google-ads-php/pull/280

examples/AdvancedOperations/AddAppCampaign.php
PierrickVoulet: I checked and we have many occurrences of ENABLED ad groups used in Shopping and Hotel code examples. I do not have strong feelings about this one as the campaign is disabled.

@AnashOommen is there any directive that needs to be followed on this?

PR #276, opened 2020-02-19
Create GitHub templates

fiboknacky:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,34 @@
...
+For general, language-agnostic questions and troubleshooting, please ask/look for answers [here](https://groups.google.com/g/adwords-api).
+
+Please check existing issues to see if your bug has already been reported before creating a new issue. We encourage you to add any relevant information to the existing issue if one already exists.
+-->
+
+**Which version of the client library are you using?**
+
+**Which version of PHP are you using?**
+
+**Which operating system (Linux, Windows, ...) and version?**
```
PierrickVoulet: @devchas what about adding a recommended way of getting these pieces of information? This would help to get accurate, complete and nicely formatted information.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,34 @@
...
+about: Report a bug in this library
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+<!--
+This form is for bug reports for this library only.
+For general, language-agnostic questions and troubleshooting, please ask/look for answers [here](https://groups.google.com/g/adwords-api).
```
PierrickVoulet: I would add "questions and troubleshooting about the **Google Ads API**". I think this is important because the group name in the provided URL is "AdWords API" which will make it confusing for people.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,34 @@
...
+
+**Actions taken**
+Please provide a code snippet or steps to consistently reproduce the experienced behavior.
+[DO NOT INCLUDE ANY PERSONAL OR SENSITIVE DATA IN YOUR SUBMITTED CODE]
+
+**Expected result**
+
+**Actual result**
+
+Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).
```
PierrickVoulet: @devchas should we add a link to the dedicated documentation we have for PHP?

.github/ISSUE_TEMPLATE/bug_report.md
devchas: For now, I think we should leave as is for the sake of brevity. If we find this becomes problematic when receiving responses, let's reconsider.

.github/ISSUE_TEMPLATE/bug_report.md
devchas: SGTM

devchas (approved):

.github/ISSUE_TEMPLATE/bug_report.md
fiboknacky: @devchas Could you confirm?

.github/ISSUE_TEMPLATE/bug_report.md
fiboknacky: I added a link to the logging page. Please let me know if you want to suggest adding other links.

.github/ISSUE_TEMPLATE/bug_report.md
PierrickVoulet: I agree it is not a blocker so will mark it as resolved but I also think that it would be useful so would love us to follow up on this later.

.github/ISSUE_TEMPLATE/bug_report.md
PierrickVoulet: Thank you both, it LGTM.

PR #280, opened 2020-02-24
AddAppCampaign fix

PierrickVoulet:
As per https://github.com/googleads/google-ads-php/pull/275#discussion_r383493678

### DIFF
--- a/examples/AdvancedOperations/AddAppCampaign.php
+++ b/examples/AdvancedOperations/AddAppCampaign.php
@@ -441,7 +441,7 @@ class AddAppCampaign
                         ])
                     ]
                     // Optional: You can set up to 20 image assets for your campaign.
-                    // 'images' => [INSERT_AD_IMAGE_ASSET_ID(s)_HERE]
+                    // 'images' => [INSERT_AD_IMAGE_RESOURCE_NAME(s)_HERE]
                 ])
             ])
         ]);


wihl (approved):

examples/AdvancedOperations/AddAppCampaign.php
```diff
@@ -441,7 +441,7 @@ private static function createAppAd(
                         ])
                     ]
                     // Optional: You can set up to 20 image assets for your campaign.
-                    // 'images' => [INSERT_AD_IMAGE_ASSET_ID(s)_HERE]
+                    // 'images' => [INSERT_AD_IMAGE_RESOURCE_NAME(s)_HERE]
```
devchas: In Java, you have to explicitly construct the AdImageAsset rather than passing a list of resource names. Just confirming this is not the case in PHP?

PierrickVoulet: Just confirmed that it is the same in PHP, we need to construct "AdImageAsset" objects. WDYT of changing for "INSERT_AD_IMAGE_ASSET(s)_HERE" instead?

devchas: I think it's more clear to actually construct the object as it would be constructed and pass in the resource name placeholder to the object.

PR #281, opened 2020-02-25
AddAppCampaign fix2

PierrickVoulet:
As per https://github.com/googleads/google-ads-php/pull/280/files#r383998855


wihl (approved):
LGTM. Approving this CL in no way obligates me to make the same change in Python. :) I think we've beaten this horse enough.

devchas (approved):

PR #285, opened 2020-03-04
Added Responsive Search Ads examples.

fiboknacky:
Based on https://github.com/googleads/google-ads-python/pull/223.

The difference is in GetResponsiveSearchAds:
* The check for emptiness of ResponsiveSearchAdInfo is excluded. As far as I checked, both headlines and descriptions need at least one member, so it's not possible that the ResponsiveSearchAdInfo will be empty.
* Due to the convenience of the language, the private function is implemented in a bit different way. It'll return the concatenation of assets directly, instead of returning an array of strings.

### DIFF
--- a/examples/BasicOperations/AddResponsiveSearchAd.php
+++ b/examples/BasicOperations/AddResponsiveSearchAd.php
@@ -1,0 +1,179 @@
+<?php
+
+/**
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Ads\GoogleAds\Examples\BasicOperations;
+
+require __DIR__ . '/../../vendor/autoload.php';
+
+use GetOpt\GetOpt;
+use Google\Ads\GoogleAds\Examples\Utils\ArgumentNames;
+use Google\Ads\GoogleAds\Examples\Utils\ArgumentParser;
+use Google\Ads\GoogleAds\Lib\OAuth2TokenBuilder;
+use Google\Ads\GoogleAds\Lib\V3\GoogleAdsClient;
+use Google\Ads\GoogleAds\Lib\V3\GoogleAdsClientBuilder;
+use Google\Ads\GoogleAds\Lib\V3\GoogleAdsException;
+use Google\Ads\GoogleAds\Util\V3\ResourceNames;
+use Google\Ads\GoogleAds\V3\Common\AdTextAsset;
+use Google\Ads\GoogleAds\V3\Common\ResponsiveSearchAdInfo;
+use Google\Ads\GoogleAds\V3\Enums\AdGroupAdStatusEnum\AdGroupAdStatus;
+use Google\Ads\GoogleAds\V3\Enums\ServedAssetFieldTypeEnum\ServedAssetFieldType;
+use Google\Ads\GoogleAds\V3\Errors\GoogleAdsError;
+use Google\Ads\GoogleAds\V3\Resources\Ad;
+use Google\Ads\GoogleAds\V3\Resources\AdGroupAd;
+use Google\Ads\GoogleAds\V3\Services\AdGroupAdOperation;
+use Google\ApiCore\ApiException;
+use Google\Protobuf\StringValue;
+
+/**
+ * This example adds a responsive search ad to a given ad group. To get ad groups,
+ * run GetAdGroups.php.
+ */
+class AddResponsiveSearchAd
+{
+    private const CUSTOMER_ID = 'INSERT_CUSTOMER_ID_HERE';
+    private const AD_GROUP_ID = 'INSERT_AD_GROUP_ID_HERE';
+
+    public static function main()
+    {
+        // Either pass the required parameters for this example on the command line, or insert them
+        // into the constants above.
+        $options = (new ArgumentParser())->parseCommandArguments([
+            ArgumentNames::CUSTOMER_ID => GetOpt::REQUIRED_ARGUMENT,
+            ArgumentNames::AD_GROUP_ID => GetOpt::REQUIRED_ARGUMENT
+        ]);
+
+        // Generate a refreshable OAuth2 credential for authentication.
+        $oAuth2Credential = (new OAuth2TokenBuilder())->fromFile()->build();
+
+        // Construct a Google Ads client configured from a properties file and the
+        // OAuth2 credentials above.
+        $googleAdsClient = (new GoogleAdsClientBuilder())->fromFile()
+            ->withOAuth2Credential($oAuth2Credential)
+            ->build();
+
+        try {
+            self::runExample(
+                $googleAdsClient,
+                $options[ArgumentNames::CUSTOMER_ID] ?: self::CUSTOMER_ID,
+                $options[ArgumentNames::AD_GROUP_ID] ?: self::AD_GROUP_ID
+            );
+        } catch (GoogleAdsException $googleAdsException) {
+            printf(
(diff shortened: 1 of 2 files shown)


examples/BasicOperations/GetResponsiveSearchAds.php
```diff
@@ -0,0 +1,175 @@
...
+
+/**
+ * This example gets non-removed responsive search ads in a specified ad group.
+ * To add responsive search ads, run BasicOperations/AddResponsiveSearchAd.php.
+ * To get ad groups, run BasicOperations/GetAdGroups.php.
+ */
+class GetResponsiveSearchAds
+{
+    private const CUSTOMER_ID = 'INSERT_CUSTOMER_ID_HERE';
+    // Optional: Specify an ad group ID below to restrict search to only a given ad group.
```
wihl: +1 I should probably add this as a useful comment to the golden.

examples/BasicOperations/GetResponsiveSearchAds.php
```diff
@@ -0,0 +1,175 @@
...
+            $query .= " AND ad_group.id = $adGroupId";
+        }
+
+        // Issues a search request by specifying page size.
+        $response =
+            $googleAdsServiceClient->search($customerId, $query, ['pageSize' => self::PAGE_SIZE]);
+
+        // Iterates over all rows in all pages and prints the requested field values for
+        // the responsive search ad in each row.
+        foreach ($response->iterateAllElements() as $googleAdsRow) {
```
wihl: If no rows are returned, it will fail silently. I think a useful message would be helpful and a good practice.

examples/BasicOperations/GetResponsiveSearchAds.php
```diff
@@ -0,0 +1,175 @@
...
+        }
+    }
+
+    /**
+     * Runs the example.
+     *
+     * @param GoogleAdsClient $googleAdsClient the Google Ads API client
+     * @param int $customerId the customer ID
+     * @param int|null $adGroupId the ad group ID for which responsive search ads will be retrieved.
+     *     If `null`, returns from all ad groups
```
PierrickVoulet: nit: "and by default"

examples/BasicOperations/GetResponsiveSearchAds.php
```diff
@@ -0,0 +1,175 @@
...
+     */
+    public static function runExample(
+        GoogleAdsClient $googleAdsClient,
+        int $customerId,
+        ?int $adGroupId
+    ) {
+        $googleAdsServiceClient = $googleAdsClient->getGoogleAdsServiceClient();
+        // Creates a query that retrieves responsive search ads.
+        $query =
```
AnashOommen: Can we remove this example please? This code example is similar to all the Get examples we already have, and it should suffice to explain in the guide that the users should use a particular AWQL to retrieve these ads.

examples/BasicOperations/GetResponsiveSearchAds.php
```diff
@@ -0,0 +1,175 @@
...
+     * @param int|null $adGroupId the ad group ID for which responsive search ads will be retrieved.
+     *     If `null`, returns from all ad groups
+     */
+    public static function runExample(
+        GoogleAdsClient $googleAdsClient,
+        int $customerId,
+        ?int $adGroupId
+    ) {
+        $googleAdsServiceClient = $googleAdsClient->getGoogleAdsServiceClient();
+        // Creates a query that retrieves responsive search ads.
```
PierrickVoulet: nit: would add a line break before the comment for readability

AnashOommen (changes requested):
Left a comment.

examples/BasicOperations/GetResponsiveSearchAds.php
```diff
@@ -0,0 +1,175 @@
...
+        $query =
+            'SELECT ad_group.id, '
+            . 'ad_group_ad.ad.id, '
+            . 'ad_group_ad.ad.responsive_search_ad.headlines, '
+            . 'ad_group_ad.ad.responsive_search_ad.descriptions, '
+            . 'ad_group_ad.status '
+            . 'FROM ad_group_ad '
+            . 'WHERE ad_group_ad.ad.type = RESPONSIVE_SEARCH_AD '
```
PierrickVoulet: nit: should it be between double quotes?

examples/BasicOperations/GetResponsiveSearchAds.php
wihl: The golden for this has been on plan for two months. Why make this change now?

fiboknacky: Anash, let's add this example. It's planned for a long time and already submitted in Perl and Python libraries.

examples/BasicOperations/AddResponsiveSearchAd.php
```diff
@@ -0,0 +1,179 @@
...
+    public static function main()
+    {
+        // Either pass the required parameters for this example on the command line, or insert them
+        // into the constants above.
+        $options = (new ArgumentParser())->parseCommandArguments([
+            ArgumentNames::CUSTOMER_ID => GetOpt::REQUIRED_ARGUMENT,
+            ArgumentNames::AD_GROUP_ID => GetOpt::REQUIRED_ARGUMENT
+        ]);
+
+        // Generate a refreshable OAuth2 credential for authentication.
```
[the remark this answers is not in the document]
fiboknacky: It's consistent with all other examples in this library, so I'm gonna keep it this way.

examples/BasicOperations/AddResponsiveSearchAd.php
```diff
@@ -0,0 +1,179 @@
...
+        $oAuth2Credential = (new OAuth2TokenBuilder())->fromFile()->build();
+
+        // Construct a Google Ads client configured from a properties file and the
```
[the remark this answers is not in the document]
fiboknacky: Ditto.

examples/BasicOperations/AddResponsiveSearchAd.php
```diff
@@ -0,0 +1,179 @@
...
+
+        // Creates an ad group ad operation.
+        $adGroupAdOperation = new AdGroupAdOperation();
+        $adGroupAdOperation->setCreate($adGroupAd);
+
+        // Issues a mutate request to add the ad group ad.
+        $adGroupAdServiceClient = $googleAdsClient->getAdGroupAdServiceClient();
+        $response = $adGroupAdServiceClient->mutateAdGroupAds($customerId, [$adGroupAdOperation]);
+
+        $createdAdGroupAdResourceName = $response->getResults()[0]->getResourceName();
```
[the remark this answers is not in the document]
fiboknacky: It's consistent with all other examples in this library, so I'm gonna keep it this way.
We make it clear that this is for a single ad. 
There are also other examples trying to add multiple items, so people can refer to those examples.

examples/BasicOperations/GetResponsiveSearchAds.php
fiboknacky: It's actually a valid grammar.
RESPONSIVE_SEARCH_AD can be thought as a literal (https://developers.google.com/google-ads/api/docs/query/grammar) and not a string.

PierrickVoulet (approved):
Only nits remaining so LGTM

examples/BasicOperations/GetResponsiveSearchAds.php
fiboknacky: Done.

examples/BasicOperations/GetResponsiveSearchAds.php
fiboknacky: I'm not sure where to place your phrase? after "If `null`"?
Anyway, it should be already covered above in declaration of AD_GROUP_ID.
The parameter $adGroupId here doesn't specify = null, so I don't think technically saying "by default" here is 100% accurate.

As this is a nit, I'm skipping for now. If we feel we should add this wording, let's do it later (but I think the more appropriate place would be in the declaration of AD_GROUP_ID or at the file-level PHPDoc)

examples/BasicOperations/GetResponsiveSearchAds.php
fiboknacky: Done.

examples/BasicOperations/GetResponsiveSearchAds.php
PierrickVoulet: There is no issue, I got confused by "?int" because I thought it mean "optional" but it does not.

PierrickVoulet (approved):

wihl (approved):

PR #295, opened 2020-03-24
Add ForecastReach code example

PierrickVoulet:


examples/Planning/ForecastReach.php
```diff
@@ -0,0 +1,389 @@
...
+{
+    const CUSTOMER_ID = 'INSERT_CUSTOMER_ID_HERE';
+
+    // US
+    const LOCATION_ID = '2840';
+
+    // 500,000
+    const BUDGET_MICROS = 500000000000;
+
+    const CURRENCY_CODE = 'USD';
```
fiboknacky: Small nit: I'd group location_id and currency_code together. It's the similar type of data.

examples/Planning/ForecastReach.php
```diff
@@ -0,0 +1,389 @@
...
+ * locations and product codes, build a media plan, and generate a video ads reach forecast.
+ */
+class ForecastReach
+{
+    const CUSTOMER_ID = 'INSERT_CUSTOMER_ID_HERE';
+
+    // US
+    const LOCATION_ID = '2840';
+
+    // 500,000
```
fiboknacky: Ditto.

examples/Planning/ForecastReach.php
```diff
@@ -0,0 +1,389 @@
...
+
+/**
+ * This example demonstrates how to interact with the ReachPlanService to find plannable
+ * locations and product codes, build a media plan, and generate a video ads reach forecast.
+ */
+class ForecastReach
+{
+    const CUSTOMER_ID = 'INSERT_CUSTOMER_ID_HERE';
+
+    // US
```
fiboknacky: Could you check if the comment at the end of the line violates PSR12?
If not, I think doing as follow is more concise and readable.

private const LOCATION_ID = '2840'; // US

(Note that private modifier is needed to comply with PSR12)

examples/Planning/ForecastReach.php
```diff
@@ -0,0 +1,389 @@
...
+        );
+
+        self::forecastSuggestedMix(
+            $googleAdsClient,
+            $customerId,
+            self::LOCATION_ID,
+            self::CURRENCY_CODE,
+            self::BUDGET_MICROS
+        );
+    }
```
fiboknacky: I think we're doing things a bit strange here.
LOCATION_ID and the other two are not open to be supplied by users.
So, why do we pass them here now? In other examples, for the real const like this, we just specify them where it's really used (e.g., PAGE_SIZE).

I don't quite get why we bother create another variables here.

fiboknacky: We should either open them to be specified by users (as optional arguments) (Preferred as it requires you to modify less, and it's more useful to users), or just remove all variables here.

examples/Planning/ForecastReach.php
```diff
@@ -0,0 +1,389 @@
...
+    private static function showPlannableLocations(GoogleAdsClient $googleAdsClient)
+    {
+        $response = $googleAdsClient->getReachPlanServiceClient()->listPlannableLocations();
+
+        printf("Plannable Locations:%sName,\tId,\tParentCountryId%s", PHP_EOL, PHP_EOL);
+        foreach ($response->getPlannableLocations() as $location) {
+            /** @var PlannableLocation $location */
+            printf(
+                "\"%s\",\t%s,\t%s%s",
+                $location->getNameUnwrapped(),
```
fiboknacky: You can use single quote inside, if you need to use double quote outside.
That's easier to read.

examples/Planning/ForecastReach.php
```diff
@@ -0,0 +1,389 @@
...
+        string $locationId
+    ) {
+        $response = $googleAdsClient->getReachPlanServiceClient()->listPlannableProducts(
+            new StringValue(['value' => $locationId])
+        );
+
+        printf('Plannable Products for Location ID %s:%s', $locationId, PHP_EOL);
+        foreach ($response->getProductMetadata() as $product) {
+            /** @var ProductMetadata $product */
+            printf("%s:%s", $product->getPlannableProductCodeUnwrapped(), PHP_EOL);
```
fiboknacky: You don't need to use printf here.
Just print like below.

examples/Planning/ForecastReach.php
```diff
@@ -0,0 +1,389 @@
...
+            print "Age Ranges:" . PHP_EOL;
```
fiboknacky: Use single quote.

fiboknacky: Same for below.

examples/Planning/ForecastReach.php
PierrickVoulet: It does not, changing for it.

`private` only fixes a warning in PSR-12. I think it is better to keep it consistent with all existing code examples (do not use `private`). WDYT of fixing this in all code examples at the same time in a separate PR instead?

examples/Planning/ForecastReach.php
PierrickVoulet: Agreed, I think the best would be to remove the useless parameters for these methods but keep the class constants for readability and flexibility (the constants are easy to find and edit when at the top of the file).

examples/Planning/ForecastReach.php
PierrickVoulet: Good point, ty.

examples/Planning/ForecastReach.php
```diff
@@ -0,0 +1,384 @@
...
+     *     by calling ListPlannableLocations on the ReachPlanService.
+     * @param string $currencyCode three-character ISO 4217 currency code
+     * @param float $budget_micros budget to allocate to the plan
+     */
+    private static function forecastManualMix(
+        GoogleAdsClient $googleAdsClient,
+        int $customerId,
+        string $locationId,
+        string $currencyCode,
+        float $budget_micros
```
fiboknacky: budgetMicros

examples/Planning/ForecastReach.php
```diff
@@ -0,0 +1,384 @@
...
+    ) {
+        // Set up a ratio to split the budget between two products.
+        $trueviewAllocation = floatval(0.15);
+        $bumperAllocation = floatval(1 - $trueviewAllocation);
+
+        // See listPlannableProducts on ReachPlanService to retrieve a list
```
fiboknacky: Nit: ListPlannableProducts

examples/Planning/ForecastReach.php
```diff
@@ -0,0 +1,384 @@
...
+    /**
+     * Pulls a forecast for a product mix based on your set of preferences.
+     *
+     * @param GoogleAdsClient $googleAdsClient the Google Ads API client
+     * @param int $customerId the customer ID
+     * @param string $locationId the location ID to plan for. You can get a valid location ID from
+     *     https://developers.google.com/adwords/api/docs/appendix/geotargeting or
+     *     by calling ListPlannableLocations on the ReachPlanService.
+     * @param string $currencyCode three-character ISO 4217 currency code
+     * @param float $budget_micros budget to allocate to the plan
```
fiboknacky: Ditto. budgetMicros
Actually, if you pass const directly, this will be gone though.

examples/Planning/ForecastReach.php
PierrickVoulet: SG, will do for now on

examples/Planning/ForecastReach.php
PierrickVoulet: Done for all methods but `requestReachCurve` because it is a utility method: I believe it makes better sense to keep it as-is.

examples/Planning/ForecastReach.php
```diff
@@ -0,0 +1,350 @@
...
+     */
+    public static function runExample(GoogleAdsClient $googleAdsClient, int $customerId)
+    {
+        self::showPlannableLocations($googleAdsClient);
+
+        self::showPlannableProducts($googleAdsClient);
+
+        self::forecastManualMix($googleAdsClient, $customerId);
+
+        self::forecastSuggestedMix($googleAdsClient, $customerId);
```
fiboknacky: I don't think we need space between these lines, unless you want to explain something (but descriptions can be found in each method already).

PR #313, opened 2020-04-09
Added UploadCallConversion example.

fiboknacky:
Based on https://github.com/googleads/google-ads-perl/pull/7/files.

### DIFF
--- a/examples/Remarketing/UploadCallConversion.php
+++ b/examples/Remarketing/UploadCallConversion.php
@@ -116,7 +116,7 @@ class UploadCallConversion
      * @param string $callStartDateTime the date and time at which the call occurred. The format is
      *     "yyyy-mm-dd hh:mm:ss+|-hh:mm", e.g. “2019-01-01 12:32:45-08:00”
      * @param string $conversionDateTime the date and time of the conversion (should be after the
-     *     click time). The format is "yyyy-mm-dd hh:mm:ss+|-hh:mm", e.g.
+     *     call time). The format is "yyyy-mm-dd hh:mm:ss+|-hh:mm", e.g.
      *     “2019-01-01 12:32:45-08:00”
      * @param float $conversionValue the value of the conversion
      */
--- a/examples/Remarketing/UploadCallConversion.php
+++ b/examples/Remarketing/UploadCallConversion.php
@@ -1,0 +1,175 @@
+<?php
+
+/**
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


examples/Remarketing/UploadCallConversion.php
```diff
@@ -0,0 +1,175 @@
...
+     *
+     * @param GoogleAdsClient $googleAdsClient the Google Ads API client
+     * @param int $customerId the customer ID
+     * @param int $conversionActionId the ID of the conversion action to upload to
+     * @param string $callerId the caller ID from which this call was placed. Caller ID is expected
+     *     to be in E.164 format with preceding '+' sign. e.g. "+16502531234"
+     * @param string $callStartDateTime the date and time at which the call occurred. The format is
+     *     "yyyy-mm-dd hh:mm:ss+|-hh:mm", e.g. “2019-01-01 12:32:45-08:00”
+     * @param string $conversionDateTime the date and time of the conversion (should be after the
+     *     click time). The format is "yyyy-mm-dd hh:mm:ss+|-hh:mm", e.g.
```
wfansh, line 119: click -> call

examples/Remarketing/UploadCallConversion.php
```diff
@@ -0,0 +1,175 @@
...
+        // Issues a request to upload the call conversion.
+        $conversionUploadServiceClient = $googleAdsClient->getConversionUploadServiceClient();
+        $response = $conversionUploadServiceClient->uploadCallConversions(
+            $customerId,
+            [$callConversion],
+            ['partialFailure' => true]
+        );
+
+        // Prints any partial errors returned.
+        if (!is_null($response->getPartialFailureError())) {
```
wfansh, line 153: I would suggest refactoring the "UploadConversionAdjustment.php" and "UploadOfflineConversion.php" to handle partial failure errors in the same approach, using "if else" statement instead of a separate method. 

Perl and Java is following the same approach
https://github.com/googleads/google-ads-java/blob/master/google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/UploadOfflineConversion.java#L149

https://github.com/googleads/google-ads-perl/blob/master/examples/remarketing/upload_offline_conversion.pl#L79

wfansh (changes requested):

examples/Remarketing/UploadCallConversion.php
fiboknacky: Why would we want to do that? Could you provide more context? 
(Actually we should discuss separately, as it's not related to this PR. I'm gonna resolve this comment anyway).

wfansh (approved):

examples/Remarketing/UploadCallConversion.php
wfansh: OK, I will send an email offline.

PierrickVoulet (approved):

PR #316, opened 2020-04-09
Rename to GetAdGroupCriterionCpcBidSimulations and fix comments and p…

PierrickVoulet:
…rints

Change-Id: I4419dedb4a4504b49d0635eef3e7fb217f4409cc

### DIFF
--- a/examples/Planning/GetAdGroupCriterionCpcBidSimulations.php
+++ b/examples/Planning/GetAdGroupCriterionCpcBidSimulations.php
@@ -34,7 +34,7 @@ use Google\Ads\GoogleAds\V3\Services\GoogleAdsRow;
 use Google\ApiCore\ApiException;
 
 /**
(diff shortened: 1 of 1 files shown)


examples/Planning/GetAdGroupCriterionCpcBidSimulations.php
```diff
@@ -34,10 +34,10 @@
 use Google\ApiCore\ApiException;
 
 /**
- * This example gets all available keyword bid simulations for a given ad group.
+ * This example gets all available CPC bid simulations for a given ad group.
```
fiboknacky, line 37: CPC along wouldn't represent the "keyword", as CPC denotes cost-per-click. I think we should include "ad group criterion" as well.
Can you include ad group criterion (like you do in the file name) all over the places?

PierrickVoulet: SG

Issue #259, opened 2020-01-10
Make all constant fields private in the code examples


PierrickVoulet:

PierrickVoulet:
Fixed.

PR #317, opened 2020-04-13
Add visibility to all code example constants

PierrickVoulet:
Change-Id: I4524c29e2bb5c6f52a449c5a7c697e6a3195719f


fiboknacky:
As a note, this will fix #259

PR #325, opened 2020-04-27
Added support of extracting customer ID from resource name in the request

fiboknacky:
References:
- https://github.com/googleads/google-ads-python/pull/262/files
- https://github.com/googleads/google-ads-ruby/pull/29

### DIFF
--- a/src/Google/Ads/GoogleAds/Lib/V1/LogMessageFormatter.php
+++ b/src/Google/Ads/GoogleAds/Lib/V1/LogMessageFormatter.php
@@ -42,23 +42,27 @@ final class LogMessageFormatter
     }
 
     /**
-     * Returns the customer ID information of the provided request.
+     * Extracts the customer ID, if present, from the provided request.
      *
      * @param Message $request the request to get its customer ID
-     * @return string the customer ID information
+     * @return string the customer ID if present or the message saying that the customer ID is not
+     *     available
      */
-    private static function getCustomerIdInfo(Message $request): string
+    private static function extractCustomerId(Message $request): string
     {
+        // Most requests contain customer ID in the request, so we aim to extract that.
         if (method_exists($request, 'getCustomerId')) {
             return $request->getCustomerId();
         } elseif (method_exists($request, 'getResourceName')) {
+            // In some cases, customer ID is available in the form of resource name, such as many
+            // Get requests.
             $resourceName = $request->getResourceName();
             $segments = explode('/', $resourceName);
             if ($segments[0] === 'customers') {
                 return $segments[1];
             }
         }
-        return '"No customer ID found in the request"';
+        return '"No customer ID could be extracted from the request"';
     }
 
     /**
@@ -89,7 +93,7 @@ final class LogMessageFormatter
             . 'IsFault: %b, FaultMessage: "%s"',
             $endpoint,
             $method,
-            self::getCustomerIdInfo($argument),
+            self::extractCustomerId($argument),
             $this->getFirstHeaderValue(self::$REQUEST_ID_HEADER_KEY, $status->metadata),
             $status->code !== 0,
             !empty($errorMessageList) ? json_encode($errorMessageList) : 'None'
--- a/src/Google/Ads/GoogleAds/Lib/V1/LogMessageFormatter.php
+++ b/src/Google/Ads/GoogleAds/Lib/V1/LogMessageFormatter.php
@@ -20,6 +20,7 @@ namespace Google\Ads\GoogleAds\Lib\V1;
 
 use Google\Ads\GoogleAds\V1\Errors\GoogleAdsFailure;
 use Google\ApiCore\ArrayTrait;
+use Google\Protobuf\Internal\Message;
 
 /**
  * Formats information about a single gRPC / REST call for logging.
@@ -38,6 +39,26 @@ final class LogMessageFormatter
     public function __construct($statusMetadataExtractor = null)
     {
         $this->statusMetadataExtractor = $statusMetadataExtractor ?: new StatusMetadataExtractor();
+    }
+
+    /**
+     * Returns the customer ID information of the provided request.
+     *
+     * @param Message $request the request to get its customer ID
+     * @return string the customer ID information
+     */
+    private static function getCustomerIdInfo(Message $request): string
+    {
+        if (method_exists($request, 'getCustomerId')) {
+            return $request->getCustomerId();
+        } elseif (method_exists($request, 'getResourceName')) {
+            $resourceName = $request->getResourceName();
+            $segments = explode('/', $resourceName);
+            if ($segments[0] === 'customers') {
+                return $segments[1];
+            }
+        }
+        return '"No customer ID found in the request"';
     }
 
     /**
@@ -64,12 +85,11 @@ final class LogMessageFormatter
         }
 
         return sprintf(
-            'Request made: Host: "%s", Method: "%s", ClientCustomerId: %d, RequestId: "%s", '
+            'Request made: Host: "%s", Method: "%s", CustomerId: %s, RequestId: "%s", '
             . 'IsFault: %b, FaultMessage: "%s"',
             $endpoint,
             $method,
-            method_exists($argument, 'getCustomerId')
-                ? $argument->getCustomerId() : '"No client customer ID used"',
+            self::getCustomerIdInfo($argument),
             $this->getFirstHeaderValue(self::$REQUEST_ID_HEADER_KEY, $status->metadata),
             $status->code !== 0,
             !empty($errorMessageList) ? json_encode($errorMessageList) : 'None'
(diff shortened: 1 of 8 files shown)


src/Google/Ads/GoogleAds/Lib/V1/LogMessageFormatter.php
```diff
@@ -40,6 +41,26 @@ public function __construct($statusMetadataExtractor = null)
         $this->statusMetadataExtractor = $statusMetadataExtractor ?: new StatusMetadataExtractor();
     }
 
+    /**
+     * Returns the customer ID information of the provided request.
+     *
+     * @param Message $request the request to get its customer ID
+     * @return string the customer ID information
+     */
+    private static function getCustomerIdInfo(Message $request): string
```
PierrickVoulet, line 50: What about `getCustomerId` or `extractCustomerId` instead?

src/Google/Ads/GoogleAds/Lib/V1/LogMessageFormatter.php
```diff
@@ -40,6 +41,26 @@ public function __construct($statusMetadataExtractor = null)
...
+    {
+        if (method_exists($request, 'getCustomerId')) {
```
PierrickVoulet, line 52: Could you add a quick comment explaining that it is covering all mutate requests?

src/Google/Ads/GoogleAds/Lib/V1/LogMessageFormatter.php
```diff
@@ -40,6 +41,26 @@ public function __construct($statusMetadataExtractor = null)
...
+            return $request->getCustomerId();
+        } elseif (method_exists($request, 'getResourceName')) {
```
PierrickVoulet, line 54: Could you add a quick comment explaining that it is covering all get requests?

src/Google/Ads/GoogleAds/Lib/V1/LogMessageFormatter.php
```diff
@@ -40,6 +41,26 @@ public function __construct($statusMetadataExtractor = null)
...
+            $resourceName = $request->getResourceName();
+            $segments = explode('/', $resourceName);
+            if ($segments[0] === 'customers') {
+                return $segments[1];
+            }
+        }
+        return '"No customer ID found in the request"';
```
PierrickVoulet, line 61: I would change for something like `No customer ID could be extracted from the request`

src/Google/Ads/GoogleAds/Lib/V1/LogMessageFormatter.php
```diff
@@ -97,7 +117,7 @@ public function formatDetail(
 
         $response = $this->pluck('response', $responseData);
         $status = $this->pluck('status', $responseData);
-        $unaryCall = $this->pluck('unaryCall', $responseData);
+        $call = $this->pluck('call', $responseData);
```
PierrickVoulet, line 120: If you want to make this change you will also need to update [here](https://github.com/googleads/google-ads-php/blob/master/src/Google/Ads/GoogleAds/Lib/V1/GoogleAdsLoggingUnaryCall.php#L57) and [here](https://github.com/googleads/google-ads-php/blob/master/src/Google/Ads/GoogleAds/Lib/V1/GoogleAdsLoggingUnaryCall.php#L61).

src/Google/Ads/GoogleAds/Lib/V1/LogMessageFormatter.php
```diff
@@ -136,7 +156,9 @@ public function formatDetail(
             ) {
                 self::$unusedGoogleAdsFailure = new GoogleAdsFailure();
             }
-            $logMessageTokens[] = "Response: {$response->serializeToJsonString()}";
+            $logMessageTokens[] = "Response: " . (is_null($response) ? "None" :
```
PierrickVoulet, line 159: I agree it is better to make it consistent with V3 implementation. No issue at all but please note that this should only be useful in the context of ServerStream (never in V1).

src/Google/Ads/GoogleAds/Lib/V1/LogMessageFormatter.php
```diff
@@ -40,6 +41,26 @@ public function __construct($statusMetadataExtractor = null)
         $this->statusMetadataExtractor = $statusMetadataExtractor ?: new StatusMetadataExtractor();
     }
 
+    /**
+     * Returns the customer ID information of the provided request.
+     *
+     * @param Message $request the request to get its customer ID
+     * @return string the customer ID information
```
PierrickVoulet, line 48: You may want to return an int instead of string (using intval)

src/Google/Ads/GoogleAds/Lib/V1/LogMessageFormatter.php
```diff
@@ -64,12 +85,11 @@ public function formatSummary(
         }
 
         return sprintf(
-            'Request made: Host: "%s", Method: "%s", ClientCustomerId: %d, RequestId: "%s", '
+            'Request made: Host: "%s", Method: "%s", CustomerId: %s, RequestId: "%s", '
```
PierrickVoulet, line 88: You can keep `%d` if you decide to return an int (see previous comment)

tests/Google/Ads/GoogleAds/Lib/V1/LogMessageFormatterTest.php
```diff
@@ -233,17 +257,48 @@ private function createSuccessRequestResponse()
...
             ->disableArgumentCloning()
             ->getMock();
-        $unaryCall->method('getMetadata')->willReturn(
+        $call->method('getMetadata')->willReturn(
             ['request-id' => ['AbCdEfGh'], 'x-google-session-info' => ['1234abcd']]
         );
 
         $this->successRequest = compact('method', 'argument', 'metadata');
-        $this->successResponse = compact('status', 'response', 'unaryCall');
+        $this->successResponse = compact('status', 'response', 'call');
```
PierrickVoulet, line 270: ditto if you want to change for `call` instead of `unaryCall` (multiple times in this file)

PierrickVoulet (changes requested):
Only commented for V1 but any changes have to be replicated to V2 & V3 accordingly.

src/Google/Ads/GoogleAds/Lib/V1/LogMessageFormatter.php
fiboknacky: Done.

src/Google/Ads/GoogleAds/Lib/V1/LogMessageFormatter.php
fiboknacky: Done.

src/Google/Ads/GoogleAds/Lib/V1/LogMessageFormatter.php
fiboknacky: We can't. We returns a string when no customer ID can be found.

src/Google/Ads/GoogleAds/Lib/V1/LogMessageFormatter.php
fiboknacky: It's not always the mutate requests. Search requests also are included here.
Listing all methods (which can be added or deleted later) will cause maintenance problems in the future.
I think Python and Perl versions are also fine. It's clear from the code itself.

src/Google/Ads/GoogleAds/Lib/V1/LogMessageFormatter.php
fiboknacky: Again, I think it's safer not to mention that explicitly. We're likely to forget to update that if anything changes.
I frame the sentence to sound like we provide an example instead.

src/Google/Ads/GoogleAds/Lib/V1/LogMessageFormatter.php
fiboknacky: No. See my reply above.

src/Google/Ads/GoogleAds/Lib/V1/LogMessageFormatter.php
fiboknacky: At first, I wanted to update only V3, but I guess you will ask me to update v2 too, so I ended up updating all versions. :)

src/Google/Ads/GoogleAds/Lib/V1/LogMessageFormatter.php
fiboknacky: Done.

tests/Google/Ads/GoogleAds/Lib/V1/LogMessageFormatterTest.php
fiboknacky: If you mean updating GoogleAdsLoggingUnaryCall, it's done. All mentioning of 'unaryCall' is changed to 'call' in this file and that of V2 already.
If not, please tell me again.

PierrickVoulet (approved):

Issue #330, opened 2020-05-01
Add an optional property to specify what transport to use


PierrickVoulet:
**Describe the problem you are trying to solve**
Have a way to configure the transport I want the client library to use instead of relying on the [default rules](https://github.com/googleapis/gax-php/blob/master/src/GapicClientTrait.php#L415) defined by gax-php: gRPC/protobuf if the C extension is loaded on the system otherwise REST/json.

**Describe the solution you would like**
Have an additional property in the client configuration ([ini file](https://github.com/googleads/google-ads-php/blob/master/examples/Authentication/google_ads_php.ini), [POPO](https://github.com/googleads/google-ads-php/blob/master/src/Google/Ads/GoogleAds/Lib/ConfigurationTrait.php#L27)) seems to be a good idea to me. This can then overload the `transport` property of the [client options](https://github.com/googleads/google-ads-php/blob/master/src/Google/Ads/GoogleAds/Lib/V3/ServiceClientFactoryTrait.php#L147) to override gax-php default rules.

**Describe alternatives you've considered**
None.

**Additional context**
I think having default rules makes total sense for simplicity and to select the most efficient transport based on the system setup.

But having a way to override these by a setting is even better: remove confusion, raise issues early if something is not configured for the given option, reduce the impact of default rule changes, testing & debugging.

PR #331, opened 2020-05-02
Add transport setting in the configuration

PierrickVoulet:
This covers the FR #330

Change-Id: I37bf02def65bd2de19d5d1d3e59d83e8b7bf9ad8

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,32 +210,9 @@ for a particular service using one of the `get...ServiceClient()` methods.
 
 ## Transport
 
-There are different types of transport that can be used. The library automatically determines which
-one is best to use by default:
-
-+ **Preferred**: [gRPC](https://grpc.io/about/). It requires the installation of the dedicated C
-extension `grpc` and it is based on HTTP/2.
-+ Alternative used when gRPC is not available: REST. It does not require any C extension 
-installation and it is based on HTTP/1.1.
-
-If you need to specify the exact transport to be used instead of relying on the default rules, you
-can set the `transport` property in the `CONNECTION` section of your `google_ads_php.ini` file:
-
-```
-[CONNECTION]
-; Optional transport settings.
-; By default, "grpc" is used if available otherwise "rest".
-transport = "grpc"
-```
-
-Alternatively, you can configure the transport setting programmatically like every other ones:
-
-```php
-$googleAdsClient = (new GoogleAdsClientBuilder())
-    ...
-    ->withTransport('grpc')
-    ->build();
-```
+There are different types of transport that can be used. See the
+[Transport guide](https://developers.google.com/google-ads/api/docs/client-libs/php/transport)
+for more information.
 
 ## Protobuf
 
--- a/README.md
+++ b/README.md
@@ -208,11 +208,39 @@ $googleAdsClient = (new GoogleAdsClientBuilder())
 Once you have an instance of `GoogleAdsClient`, you can obtain a service client
 for a particular service using one of the `get...ServiceClient()` methods.
 
+## Transport
+
+There are different types of transport that can be used. The library automatically determines which
+one is best to use by default:
+
++ **Preferred**: [gRPC](https://grpc.io/about/). It requires the installation of the dedicated C
+extension `grpc` and it is based on HTTP/2.
++ Alternative used when gRPC is not available: REST. It does not require any C extension 
+installation and it is based on HTTP/1.1.
+
+If you need to specify the exact transport to be used instead of relying on the default rules, you
+can set the `transport` property in the `CONNECTION` section of your `google_ads_php.ini` file:
+
+```
+[CONNECTION]
+; Optional transport settings.
+; By default, "grpc" is used if available otherwise "rest".
+transport = "grpc"
+```
+
+Alternatively, you can configure the transport setting programmatically like every other ones:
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -208,11 +208,39 @@ $googleAdsClient = (new GoogleAdsClientBuilder())
 Once you have an instance of `GoogleAdsClient`, you can obtain a service client
 for a particular service using one of the `get...ServiceClient()` methods.
 
+## Transport
```
fiboknacky, line 211: Provided that this is a nice-to-have feature, we should put this section in [DevSite](https://developers.google.com/google-ads/api/docs/client-libs/php/) instead.
Then, you can just briefly mention above (around the **Requirement** section) that the property file can be configured to specify transport.

tests/Google/Ads/GoogleAds/Lib/V3/GoogleAdsClientBuilderTest.php
```diff
@@ -80,6 +82,15 @@ public function testBuildClientFromConfiguration()
         $this->assertSame(self::$LOGIN_CUSTOMER_ID, $googleAdsClient->getLoginCustomerId());
         $this->assertSame('https://abc.xyz:443', $googleAdsClient->getEndpoint());
         $this->assertSame('https://localhost:8080', $googleAdsClient->getProxy());
+        $this->assertSame(self::provideValidTransports()[0][0], $googleAdsClient->getTransport());
+    }
+
+    public function provideValidTransports()
```
fiboknacky, line 88: 1. Is this needed to be public? Will `@dataProvider` work if this is private? (I know it would not be consistent with other methods, but we should do what is correct)
2. Can this be static? If not, using `$this->provideValidTransports` seems more semantically correct.

tests/Google/Ads/GoogleAds/Lib/V3/GoogleAdsClientBuilderTest.php
```diff
@@ -62,7 +63,8 @@ public function testBuildClientFromConfiguration()
             ['developerToken', 'GOOGLE_ADS', self::$DEVELOPER_TOKEN],
             ['loginCustomerId', 'GOOGLE_ADS', self::$LOGIN_CUSTOMER_ID],
             ['endpoint', 'GOOGLE_ADS', 'https://abc.xyz:443'],
-            ['proxy', 'CONNECTION', 'https://localhost:8080']
+            ['proxy', 'CONNECTION', 'https://localhost:8080'],
+            ['transport', 'CONNECTION', self::provideValidTransports()[0][0]]
```
fiboknacky, line 67: This is better just 'grpc'.
We don't benefit from using `provideValidTransports` here, so using it make it harder to read this code, especially with double indices [0].

tests/Google/Ads/GoogleAds/Lib/V3/GoogleAdsClientBuilderTest.php
```diff
@@ -80,6 +82,15 @@ public function testBuildClientFromConfiguration()
         $this->assertSame(self::$LOGIN_CUSTOMER_ID, $googleAdsClient->getLoginCustomerId());
         $this->assertSame('https://abc.xyz:443', $googleAdsClient->getEndpoint());
         $this->assertSame('https://localhost:8080', $googleAdsClient->getProxy());
+        $this->assertSame(self::provideValidTransports()[0][0], $googleAdsClient->getTransport());
```
fiboknacky, line 85: Ditto

tests/Google/Ads/GoogleAds/Lib/V3/GoogleAdsClientBuilderTest.php
fiboknacky, line 88: You can move this down to close to `provideInvalidProxyURIs()`. That's why we use this.

README.md
PierrickVoulet: Agreed, the documentation is transferred to a new guide: https://developers.google.com/google-ads/api/docs/client-libs/php/transport.

tests/Google/Ads/GoogleAds/Lib/V3/GoogleAdsClientBuilderTest.php
PierrickVoulet: 1. Yes it does: see doc [here](https://phpunit.de/manual/3.7/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers)
1. Yes it can be static so I changed.

tests/Google/Ads/GoogleAds/Lib/V3/GoogleAdsClientBuilderTest.php
```diff
@@ -309,6 +301,14 @@ public function testBuildWithValidTransports(string $transport)
         $this->assertSame($transport, $googleAdsClient->getTransport());
     }
 
+    static public function provideValidTransports()
```
fiboknacky, line 304: Nit: Let's use `public static` for consistency.

PierrickVoulet: Good catch, it is required to be PSR-12 compliant!

PR #334, opened 2020-05-05
Add failure exit codes for all code examples

PierrickVoulet:
Make all code examples return:
- 0 no failure
- 1 when API failure
- 255 when uncaught failure

Change-Id: Ib991eba94208b5e804166a5fd7bb287d0ffaada8

### DIFF
--- a/examples/AccountManagement/ApproveMerchantCenterLink.php
+++ b/examples/AccountManagement/ApproveMerchantCenterLink.php
@@ -91,12 +91,14 @@ class ApproveMerchantCenterLink
                     PHP_EOL
                 );
             }
+            exit(1);
         } catch (ApiException $apiException) {
             printf(
                 "ApiException was thrown with message '%s'.%s",
                 $apiException->getMessage(),
                 PHP_EOL
             );
+            exit(1);
         }
     }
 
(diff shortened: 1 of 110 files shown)


PierrickVoulet:
> Can you pls wait on this PR until Fan comes with a clear guideline on what to do?
> 
> Two things to consider.
> 
> 1. Will you ever write your own code example tests? If yes, will phpunit like the exit calls?
> 2. Are the error codes you picked the ones that Fan needs?

1. I cannot think of any benefit of having UTs for a code example, WDYT @fiboknacky ?
2. I believe I did, these had been decided a few weeks back, do you agree @wfansh ?

fiboknacky:
> 1. I cannot think of any benefit of having UTs for a code example, WDYT @fiboknacky ?

I don't think UT makes sense here too. Even if we want to implement the example tests in the future, I don't see why they'd be a blocker of this effort

> I believe I did, these had been decided a few weeks back, do you agree @wfansh ?

I understand the same as Pierrick. Hasn't it been decided?

wfansh:
I think we've agreed to use 0 for code examples success, 1 for API errors and 255 for other errors.

On my side, I will check whether the return code is 0 for all languages.

PierrickVoulet:
> I think we've agreed to use 0 for code examples success, 1 for API errors and 255 for other errors.
> 
> On my side, I will check whether the return code is 0 for all languages.

Thank you for confirming @wfansh ! @AnashOommen any strong feelings before I merge ?

PR #344, opened 2020-05-20
Improve the HandleKeywordPolicyViolations code example

### LINKED ISSUE #340: Refactor HandleKeywordPolicyViolations example?
PTAL https://github.com/googleads/google-ads-dotnet/pull/186 where I refactored HandleKeywordPolicyViolations example a bit to 

1) Rethrow exceptions like in HandleExpandedTextAdsPolicyViolations
2) Keep track of non-exemptable errors (instead of making assumptions on error count) to determine if a policy exemption can be requested.

Could you PTAL and see if we can get those improvements in the PHP example?

PierrickVoulet:
The improvement is based on the [implementation](https://github.com/googleads/google-ads-dotnet/pull/186) in progress in DotNet.

Fixed #340 

Change-Id: Iee0761f7effdb568fd1e0680a900d653a649a677

### DIFF
--- a/examples/ErrorHandling/HandleKeywordPolicyViolations.php
+++ b/examples/ErrorHandling/HandleKeywordPolicyViolations.php
@@ -228,10 +228,14 @@ class HandleKeywordPolicyViolations
                         PHP_EOL
                     );
                 } else {
-                    print "No exemption request is sent because your keyword contained some"
-                        . " non-exemptible policy violations." . PHP_EOL;
+                    print "No exemption request is sent because your keyword contained some "
+                        . "non-exemptible policy violations." . PHP_EOL;
                     throw $googleAdsException;
                 }
+            } else {
+                print "No exemption request is sent because there are other non-policy related "
+                    . "errors thrown." . PHP_EOL;
+                throw $googleAdsException;
             }
         }
         return $exemptPolicyViolationKeys;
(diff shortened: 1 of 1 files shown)


examples/ErrorHandling/HandleKeywordPolicyViolations.php
```diff
@@ -246,10 +240,12 @@ private static function fetchExemptPolicyViolationKeys(GoogleAdsException $googl
     /**
      * Sends exemption requests for creating a keyword.
      *
-     * @param int $customerId
-     * @param AdGroupCriterionServiceClient $adGroupCriterionServiceClient
-     * @param AdGroupCriterionOperation $adGroupCriterionOperation
-     * @param PolicyViolationKey[] $exemptPolicyViolationKeys
+     * @param int $customerId the customer ID
+     * @param AdGroupCriterionServiceClient $adGroupCriterionServiceClient the Ad Group Criterion
```
fiboknacky, line 244: Either "the ad group criterion service" or "AdGroupCriterionService"

examples/ErrorHandling/HandleKeywordPolicyViolations.php
```diff
@@ -237,6 +227,10 @@ private static function fetchExemptPolicyViolationKeys(GoogleAdsException $googl
                         $policyViolationDetailsKey->getViolatingTextUnwrapped(),
                         PHP_EOL
                     );
+                } else {
+                    print "No exemption request is sent because your keyword contained some"
+                        . " non-exemptible policy violations." . PHP_EOL;
+                    throw $googleAdsException;
                 }
```
fiboknacky, line 234: Is this message actually accurate? It's just for one error.
We still have other error in this for loop to check.
If there is at least one error that has  exemptible policy violation, `$exemptPolicyViolationKeys` is still added, and the policy exemption request will still be sent.

PierrickVoulet: Good catch!

PierrickVoulet:
> The new behavior is to also throw exceptions when we find some errors are not exemptible, correct?

Yes, the intent is to bubble up the exception if the associated errors are not adequate to send the exemption. This includes the two scenarios: 

- The keyword contains some non-exemptible policy violations
- There are other non-policy related errors

@AnashOommen feel free to chime in if I am wrong

PR #345, opened 2020-05-26
Added AddSitelinksUsingFeeds example

fiboknacky:


PierrickVoulet (changes requested):

examples/Extensions/AddSitelinksUsingFeeds.php
```diff
@@ -0,0 +1,531 @@
...
+        GoogleAdsClient $googleAdsClient,
+        int $customerId,
+        int $campaignId,
+        ?int $adGroupId
+    ) {
+        // Create a feed, which acts as a table to store data.
+        $feedData = self::createFeed($googleAdsClient, $customerId);
+
+        // Create feed items, which fill out the feed table with data.
+        $feedData = self::createFeedItems($googleAdsClient, $customerId, $feedData);
```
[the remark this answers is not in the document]
fiboknacky: I know we can pass it by reference, but I'm quite neutral to adopting that style for this case because of some reasons:
- We still have `createFeed` that returns the feed data as an output. Doing the same here is clear to me in terms of the data flow. We pass feed data in and we get new data back. This style is somewhat adopted by some PHP native functions.
- Our function is not complicated enough that we need passing by reference to help. For instance, if the function needed to return something else as output, I would pass this parameter as reference too, to avoid create a complex object just to stuff those output data together.

examples/Extensions/AddSitelinksUsingFeeds.php
```diff
@@ -0,0 +1,529 @@
...
+     * @param int $campaignId the campaign ID
+     * @param int|null $adGroupId the ad group ID
+     */
+    public static function runExample(
+        GoogleAdsClient $googleAdsClient,
+        int $customerId,
+        int $campaignId,
+        ?int $adGroupId
+    ) {
+        // Create a feed, which acts as a table to store data.
```
wfansh, line 132: The convention for Perl is to use 3rd-person verbs for method annotation but 1st-person verbs for regular comments within the method. (inherit from AdWords API library).

But for PHP, I know we're using 3rd-person verbs for all. Do you mind changing  "Create" to "Creates"?

examples/Extensions/AddSitelinksUsingFeeds.php
```diff
@@ -0,0 +1,529 @@
...
+        $feedData = self::createFeed($googleAdsClient, $customerId);
+
+        // Create feed items, which fill out the feed table with data.
```
wfansh, line 135: Create - Creates

examples/Extensions/AddSitelinksUsingFeeds.php
```diff
@@ -0,0 +1,529 @@
...
+        $feedData = self::createFeedItems($googleAdsClient, $customerId, $feedData);
+
+        // Create a feed mapping, which tells Google Ads how to interpret this data to
```
wfansh, line 138: 3rd person verbs as ablve.

examples/Extensions/AddSitelinksUsingFeeds.php
```diff
@@ -0,0 +1,529 @@
...
+        // display additional sitelink information on ads.
+        self::createFeedMapping($googleAdsClient, $customerId, $feedData);
+
+        // Create a campaign feed, which tells Google Ads which campaigns to use the
```
wfansh, line 142: 3rd person verbs as ablve.

examples/Extensions/AddSitelinksUsingFeeds.php
```diff
@@ -0,0 +1,529 @@
...
+     * @return array the feed data containing the created feed resource name and the feed attribute
+     *     IDs
+     */
+    private static function createFeed(GoogleAdsClient $googleAdsClient, int $customerId)
+    {
+        // Creates a feed.
+        $feed = new Feed([
+            'name' => new StringValue(['value' => 'Sitelinks Feed #' . uniqid()]),
+            'origin' => FeedOrigin::USER,
+            // Specify the column name and data type. This is just raw data at this point,
```
wfansh, line 166: 3rd person verbs as ablve.

wfansh (changes requested):

wfansh (approved):

examples/Extensions/AddSitelinksUsingFeeds.php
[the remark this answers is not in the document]
PierrickVoulet: SGTM

PierrickVoulet (approved):

PR #348, opened 2020-05-28
Upgrade JSON content to be compliant to the Protobuf spec

PierrickVoulet:
Change-Id: Iee632d505153db61a2488f7f41e1dfdd18b71640

### DIFF
--- a/src/Google/Ads/GoogleAds/Util/Testing/testdata/test_cases.json
+++ b/src/Google/Ads/GoogleAds/Util/Testing/testdata/test_cases.json
@@ -1,112 +1,112 @@
 {
-  "test_cases": [{
+  "testCases": [{
     "description": "Modify scalar in a wrapper proto",
-    "original_resource": {
+    "originalResource": {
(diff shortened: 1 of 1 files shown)


src/Google/Ads/GoogleAds/Util/Testing/testdata/test_cases.json
```diff
@@ -1,112 +1,112 @@
...
+    "modifiedResource": {
       "foo": {
         "bar": {
-          "nested_wrapper": true
+          "nestedWrapper": true
         }
       }
     },
-    "expected_mask": "foo.bar.nested_wrapper"
+    "expectedMask": "foo.bar.nestedWrapper"
```
fiboknacky, line 110: The generated source still has snake_cased field names. Would this work?

PierrickVoulet: I double checked, it works just fine.

PR #349, opened 2020-05-28
Fix UT asserts

PierrickVoulet:
Asserts always return true when comparing two FieldMask instances. I switch to a comparison of their JSON string representation instead.

KNOWN ISSUE: 2 UTs fail with this new way of assertion, I mark this PR as DRAFT for now until a fix is found.

Change-Id: I3b890a3bdcceb21876b157eda96d066938638387

### DIFF
--- a/tests/Google/Ads/GoogleAds/Util/FieldMasksTest.php
+++ b/tests/Google/Ads/GoogleAds/Util/FieldMasksTest.php
@@ -54,9 +54,25 @@ class FieldMasksTest extends TestCase
     public function testFieldMaskCompare(
         $originalResource,
         $modifiedResource,
-        FieldMask $expectedFieldMask
+        FieldMask $expectedFieldMask,
+        string $description
     ) {
         $actualFieldMask = FieldMasks::compare($originalResource, $modifiedResource);
+        print "============ TEST '" . $description . "'" .  PHP_EOL;
+        print "- expected: ";
+        foreach ($expectedFieldMask->getPaths() as $path) {
+            print $path . " ";
+        };
+        print PHP_EOL . "- actual: ";
+        foreach ($actualFieldMask->getPaths() as $path) {
+            print $path . " ";
+        };
+        print PHP_EOL . "Testing with FieldMask instances...";
+        $this->assertEquals(
+            $expectedFieldMask,
+            $actualFieldMask
+        );
+        print PHP_EOL . "Testing with FieldMask JSON string...";
         $this->assertEquals(
             $expectedFieldMask->serializeToJsonString(),
             $actualFieldMask->serializeToJsonString()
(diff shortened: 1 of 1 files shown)


PierrickVoulet:
@fiboknacky I see you changed some UTs in #291, any further context you could provide before I deep dive into the FieldMarks util class by any chance?

fiboknacky:
> @fiboknacky I see you changed some UTs in #291, any further context you could provide before I deep dive into the FieldMarks util class by any chance?

I'm not sure what in particular you would like me to provide more. Could you elaborate?
I think all the context was discussed and given during the review process.

To provide a summary here again, the PR was to fix the wrong behavior of FieldMasks. This makes the PHP lib aligned with the Java lib as well.
I'd recommend comparing the "before" and "after" versions of `test_cases.json`. That would show what specific cases we wanted to fix.

PierrickVoulet:
> > Asserts always return true when comparing two FieldMask instances. I switch to a comparison of their JSON string representation instead.
> 
> What do you mean by this? Do you mean `testFieldMaskCompare` always passed before?

What I mean is that the assertion of two FieldMask instances always returns `true` even when they are not equal.

```php
<?php
// ...
$this->assertEquals($expectedFieldMask, $actualFieldMask);
```

PierrickVoulet:
@fiboknacky sorry I forgot to flag the PR as draft, this code is exploratory. I added you as reviewer but I only wanted to gather historical context about these tests.

**Summary**
This PR aims to fix two distinct issues:
1. The implementation of the FieldMask UTs which does not work as expected: the asserts always return `true` even when they should return `false`
1. Fixing the implementation of the FieldMask UTs, I noticed that two UTs that are related to repeated field modification fail.

**Details**
I updated the code to better highlight these two issues we are facing. Here are the two UTs that do not fail using the current implementation but do when using my not-perfect-but-working implementation that compares the paths as JSON strings instead.

```
============ TEST 'Modify element in a repeated field'
- expected: foos 
- actual: 
Testing with FieldMask instances...
Testing with FieldMask JSON string...
Failed asserting that two strings are equal.
Expected :'"foos"'
Actual   :'""'
```

```
============ TEST 'Modify element in a nested repeated field'
- expected: foos 
- actual: 
Testing with FieldMask instances...
Testing with FieldMask JSON string...
Failed asserting that two strings are equal.
Expected :'"foos"'
Actual   :'""'
```

fiboknacky:
Understood. Please let me know when it's ready for review.

PierrickVoulet:
Quick update: after further investigation the UTs fail due to wrong comparison of repeated fields. I opened the issue https://github.com/haberman/protobuf/issues/13 to confirm.

PierrickVoulet:
The current behavior of repeated field is a bug, fix should come from this https://github.com/protocolbuffers/protobuf/issues/7650.

PierrickVoulet:
I just tested with protocolbuffers/protobuf#7650 and protocolbuffers/protobuf#7900 and it seems to be fixed. Let's wait for the next protobuf release before closing this though.

PierrickVoulet:
I confirm that all the unit tests pass successfully with the new release of `protobuf`, I am closing this because it is fixed.

PR #354, opened 2020-06-05
Fix generate keyword ideas

PierrickVoulet:
1. Be consistent with multiple argument names
1. Fix issue when optional argument is not provided (null value by default)


examples/Utils/ArgumentNames.php
```diff
@@ -65,12 +65,14 @@ final class ArgumentNames
     public const HOTEL_CENTER_ACCOUNT_ID = 'hotelCenterAccountId';
     public const KEYWORD_PLAN_ID = 'keywordPlanId';
     public const KEYWORD_TEXT = 'keywordText';
```
fiboknacky, line 67: The old ones are still used by other examples, right?

PierrickVoulet: Both argument names are used elsewhere in other code examples so I kept them.

PR #357, opened 2020-06-09
Added tests for ServiceClientFactoryTrait

fiboknacky:
Tests are generated by `preg_replace` based on the import statements of each versioned `ServiceClientFactoryTrait`.


PierrickVoulet:
@fiboknacky I understand that you use a script for this? If it is the case, any chance to add it with this PR for future regens?

PierrickVoulet (commented):
I understand that you use a script for this? If it is the case, any chance to add it with this PR for future regens?

fiboknacky:
> I understand that you use a script for this? If it is the case, any chance to add it with this PR for future regens?

It's a semi-auto process. It's not impossible to add that somewhere, e.g., as a part of our releasing process, but it's not ready yet.
I'll track that separatel.

PierrickVoulet (approved):

PR #359, opened 2020-06-10
Added tests for StatusMetadataExtractor.

fiboknacky:
The content for all versions is the same.

### DIFF
--- a/tests/Google/Ads/GoogleAds/Lib/V1/StatusMetadataExtractorTest.php
+++ b/tests/Google/Ads/GoogleAds/Lib/V1/StatusMetadataExtractorTest.php
@@ -36,18 +36,17 @@ class StatusMetadataExtractorTest extends TestCase
     /** @var StatusMetadataExtractor $statusMetadataExtractor */
     private $statusMetadataExtractor;
 
+    /** @var array $defaultStatusMetadata */
+    private $defaultStatusMetadata;
+
     /**
      * @see \PHPUnit\Framework\TestCase::setUp()
      */
     protected function setUp()
     {
         $this->statusMetadataExtractor = new StatusMetadataExtractor();
-    }
-
-    public function testExtractGoogleAdsFailure()
-    {
-        $expected = new GoogleAdsFailure();
-        $expected->setErrors([
+        $defaultGoogleAdsFailure = new GoogleAdsFailure();
+        $defaultGoogleAdsFailure->setErrors([
             new GoogleAdsError([
                 'message' => 'failure 1',
                 'error_code' => new ErrorCode([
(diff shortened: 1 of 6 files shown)


tests/Google/Ads/GoogleAds/Lib/V1/StatusMetadataExtractorTest.php
```diff
@@ -0,0 +1,166 @@
...
+     * @see \PHPUnit\Framework\TestCase::setUp()
+     */
+    protected function setUp()
+    {
+        $this->statusMetadataExtractor = new StatusMetadataExtractor();
+    }
+
+    public function testExtractGoogleAdsFailure()
+    {
+        $expected = new GoogleAdsFailure();
```
PierrickVoulet, line 49: The creation of this object is done at least two times in the tests (`testExtractGoogleAdsFailure ` and `testExtractErrorMessageList `), could you factorize?

Ditto in V2 and V3.

PierrickVoulet (changes requested):
We got many issues reported about the extraction of errors. Do these tests cover it all?

Also, IIRC there is a big difference in format whether REST or gRPC is used, do we want to add tests specifically for REST here?

fiboknacky:
> We got many issues reported about the extraction of errors. Do these tests cover it all?
> 
> Also, IIRC there is a big difference in format whether REST or gRPC is used, do we want to add tests specifically for REST here?

No. It's not an aim of this PR. This PR is just for adding tests of existing classes.

PierrickVoulet:
> > We got many issues reported about the extraction of errors. Do these tests cover it all?
> > Also, IIRC there is a big difference in format whether REST or gRPC is used, do we want to add tests specifically for REST here?
> 
> No. It's not an aim of this PR. This PR is just for adding tests of existing classes.

Thanks for confirming, we are good then. I let you open issue(s) if you wish so that we can prioritize what is not in the scope of this PR later.

PierrickVoulet (approved):

fiboknacky:
> > > We got many issues reported about the extraction of errors. Do these tests cover it all?
> > > Also, IIRC there is a big difference in format whether REST or gRPC is used, do we want to add tests specifically for REST here?
> > 
> > 
> > No. It's not an aim of this PR. This PR is just for adding tests of existing classes.
> 
> Thanks for confirming, we are good then. I let you open issue(s) if you wish so that we can prioritize what is not in the scope of this PR later.

In fact, I quickly check the code again and it seems everything is already in place. It allows for extracting both grpc and REST keys.
Based on the previous discussions, I found most complaints were usually based on misunderstanding in some ways.

PR #362, opened 2020-06-11
Rename BID_MICRO_AMOUNT to CPC_BID_MICRO_AMOUNT in UpdateAdGroup

PierrickVoulet:
Change-Id: Ied0c5f819bf7955c0dc0e4809cba607af1ae6e1a


PierrickVoulet:
> We changed it back? Or I just misunderstand.

This is to fix b/158738822

PR #368, opened 2020-06-18
Added AddDisplayUploadAd example.

fiboknacky:


examples/AdvancedOperations/AddDisplayUploadAd.php
```diff
@@ -0,0 +1,243 @@
...
+require __DIR__ . '/../../vendor/autoload.php';
+
+use GetOpt\GetOpt;
+use Google\Ads\GoogleAds\Examples\Utils\ArgumentNames;
+use Google\Ads\GoogleAds\Examples\Utils\ArgumentParser;
+use Google\Ads\GoogleAds\Lib\OAuth2TokenBuilder;
+use Google\Ads\GoogleAds\Lib\V3\GoogleAdsClient;
+use Google\Ads\GoogleAds\Lib\V3\GoogleAdsClientBuilder;
+use Google\Ads\GoogleAds\Lib\V3\GoogleAdsException;
+use Google\Ads\GoogleAds\Util\V1\ResourceNames;
```
wfansh, line 30: should this be V3?

examples/AdvancedOperations/AddDisplayUploadAd.php
```diff
@@ -0,0 +1,243 @@
...
+            ]),
+            'ad_group' => ResourceNames::forAdGroup($customerId, $adGroupId),
+            'status' => AdGroupAdStatus::PAUSED
+        ]);
+
+        // Creates an ad group ad operation.
+        $adGroupAdOperation = new AdGroupAdOperation();
+        $adGroupAdOperation->setCreate($adGroupAd);
+
+        // Issues a mutate request to add the ad group ads.
```
wfansh, line 229: "ads" to "ad"?

https://github.com/googleads/google-ads-php/blob/master/examples/BasicOperations/AddResponsiveSearchAd.php#L152

wfansh (changes requested):

examples/AdvancedOperations/AddDisplayUploadAd.php
```diff
@@ -0,0 +1,243 @@
...
+{
+    private const CUSTOMER_ID = 'INSERT_CUSTOMER_ID_HERE';
+    private const AD_GROUP_ID = 'INSERT_AD_GROUP_ID_HERE';
+
+    public static function main()
+    {
+        // Either pass the required parameters for this example on the command line, or insert them
+        // into the constants above.
+        $options = (new ArgumentParser())->parseCommandArguments(
+            [
```
PierrickVoulet, line 63: nit: you could remove the extra indentation for `[` and `]`, see how [GetCampaigns.php](https://github.com/googleads/google-ads-php/blob/master/examples/BasicOperations/GetCampaigns.php#L44) does for reference.

I noticed a few other occurrences later in the code. This is PSR-12 compliant so not a problem if you prefer to keep it that way.

examples/AdvancedOperations/AddDisplayUploadAd.php
```diff
@@ -0,0 +1,243 @@
...
+     */
+    public static function runExample(
+        GoogleAdsClient $googleAdsClient,
+        int $customerId,
+        int $adGroupId
+    ) {
+        // There are several types of display upload ads. For this example, we will create
+        // an HTML5 upload ad, which requires a media bundle.
+        // The DisplayUploadProductType field lists the available display upload types:
+        // https://developers.google.com/google-ads/api/reference/rpc/v3/DisplayUploadAdInfo
```
PierrickVoulet, line 126: Related to previous comment: should not this be https://developers.google.com/google-ads/api/reference/rpc/v3/DisplayUploadProductTypeEnum.DisplayUploadProductType?

examples/AdvancedOperations/AddDisplayUploadAd.php
```diff
@@ -0,0 +1,243 @@
...
+     * @param int $adGroupId the ad group ID to add a display upload ad to
+     */
+    public static function runExample(
+        GoogleAdsClient $googleAdsClient,
+        int $customerId,
+        int $adGroupId
+    ) {
+        // There are several types of display upload ads. For this example, we will create
+        // an HTML5 upload ad, which requires a media bundle.
+        // The DisplayUploadProductType field lists the available display upload types:
```
PierrickVoulet, line 125: I find this sentence misleading because it mixes `DisplayUploadProductType`, `DisplayUploadAdInfo` and `DisplayUploadAdInfo.display_upload_product_type`. What about only referencing `DisplayUploadProductType` here? More details about `DisplayUploadAdInfo` could then be provided in the  `createDisplayUploadAdGroupAd` function

examples/AdvancedOperations/AddDisplayUploadAd.php
```diff
@@ -0,0 +1,243 @@
...
+            ]
+        );
+
+        // Creates an asset operation.
+        $assetOperation = new AssetOperation();
+        $assetOperation->setCreate($asset);
+
+        // Issues a mutate request to add the asset.
+        $assetServiceClient = $googleAdsClient->getAssetServiceClient();
+        $response = $assetServiceClient->mutateAssets(
```
PierrickVoulet, line 174: This fits in 1 line

examples/AdvancedOperations/AddDisplayUploadAd.php
```diff
@@ -0,0 +1,243 @@
...
+            $customerId,
+            [$assetOperation]
+        );
+
+        // Prints the resource name of the added media bundle asset.
+        /** @var MutateAssetResult $addedMediaBundleAsset */
+        $addedMediaBundleAsset = $response->getResults()[0];
```
PierrickVoulet, line 181: nit: you may want to use a variable for the resource name here, something similar to `$adGroupAdResourceName = $adGroupAdResponse->getResults()[0]->getResourceName();`

This is meaningful here because you request twice the `getResourceName()` method

PierrickVoulet (changes requested):

examples/AdvancedOperations/AddDisplayUploadAd.php
wfansh: I think this's following golden CL.

examples/AdvancedOperations/AddDisplayUploadAd.php
wfansh: I think this's following golden CL :)

wfansh (changes requested):

examples/AdvancedOperations/AddDisplayUploadAd.php
fiboknacky: Done. Thanks for catching

examples/AdvancedOperations/AddDisplayUploadAd.php
fiboknacky: It's hard to make the code formatter fix in this way.
I'll fix for this time but  will probably need to leave it as is sometimes for faster implementation.

examples/AdvancedOperations/AddDisplayUploadAd.php
fiboknacky: I don't see why it's a problem here.
If a user follows the link, the user can see DisplayUploadProductType in there.
We can also educate users about both DisplayUploadProductType and DisplayUploadAdInfo at the same time, so I see this as advantage.

examples/AdvancedOperations/AddDisplayUploadAd.php
fiboknacky: Ditto.

wfansh (approved):

examples/AdvancedOperations/AddDisplayUploadAd.php
PierrickVoulet: SGTM

examples/AdvancedOperations/AddDisplayUploadAd.php
PierrickVoulet: SGTM

PierrickVoulet (approved):

PR #382, opened 2020-07-14
Add AddBillingSetup code example

PierrickVoulet:
Golden: https://github.com/googleads/google-ads-dotnet/blob/master/examples/Billing/AddBillingSetup.cs

Change-Id: I0398196e2d9411685058bd2c57941cabc35ad130

### DIFF
--- a/examples/Billing/AddBillingSetup.php
+++ b/examples/Billing/AddBillingSetup.php
@@ -40,8 +40,10 @@ use Exception;
 
 /**
  * This example creates a billing setup for a customer. A billing setup is a link between
- * a payment accounts and a customer. The new billing setup can either reuse an existing payments
+ * a payments account and a customer. The new billing setup can either reuse an existing payments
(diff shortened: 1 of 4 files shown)


examples/Billing/AddBillingSetup.php
```diff
@@ -198,8 +198,8 @@ public static function createBillingSetup(
                 'payments_profile_id' => new StringValue(['value' => $paymentsProfileId])
             ]));
         } else {
-            throw new Exception(
-                'No payments account ID or payments profile ID provided.' . PHP_EOL
+            throw new \RuntimeException(
```
fiboknacky, line 201: WDYT about UnexpectedValueException?
Should be better than RuntimeException.

PierrickVoulet: SGTM, done

PierrickVoulet:
> Please add a test for new method in ResourceNames.

Added.

PR #384, opened 2020-07-14
Add AddLogicalUserList code example

PierrickVoulet:
Change-Id: Ifd416b1bff7157658397eb05295d7fb92532b2d9

### DIFF
--- a/examples/Remarketing/AddUnionUserList.php
+++ b/examples/Remarketing/AddUnionUserList.php
@@ -152,7 +152,8 @@ class AddUnionUserList
         // Issues a mutate request to add the user list and prints some information.
         $userListServiceClient = $googleAdsClient->getUserListServiceClient();
         $response = $userListServiceClient->mutateUserLists($customerId, [$operation]);
-        printf("Created combination user list with resource name '%s'.%s",
+        printf(
+            "Created combination user list with resource name '%s'.%s",
             $response->getResults()[0]->getResourceName(),
             PHP_EOL
         );
--- a/examples/Remarketing/AddUnionUserList.php
+++ b/examples/Remarketing/AddUnionUserList.php
@@ -1,0 +1,162 @@
+<?php
+
+/**
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


examples/Remarketing/AddUnionUserList.php
```diff
@@ -0,0 +1,163 @@
...
+use Google\Ads\GoogleAds\V4\Resources\UserList;
+use Google\Ads\GoogleAds\V4\Services\UserListOperation;
+use Google\ApiCore\ApiException;
+use Google\Protobuf\StringValue;
+
+/**
+ * Creates a combination user list containing users that are present on any one of the provided user
+ * lists.
+ */
+class AddUnionUserList
```
fiboknacky, line 45: I wonder why we name this "UnionUserList".
Nowhere in the code that has the word "union". 
Instead, we call this kind of user list "combination user list" in many places.

PierrickVoulet: @devchas any background you could provide on this as you wrote the golden?

devchas: Please see the [conversation history](https://github.com/googleads/google-ads-java/pull/241) on the PR - specifically comments made by Nick on Mar 3.

PierrickVoulet: Thanks @devchas for the link. SGTM, any strong feelings about this @fiboknacky ?

PierrickVoulet: @fiboknacky Is it OK with you if I keep it like that?

fiboknacky: If Devin doesn't object to this, I prefer we use the name CombinationUserList instead.
We use that word in so many places throughout the code, so why do we just name the example differently?

devchas: I am okay with changing the name of the example. However, I think AddCombinationUserList is too similar to AddCombinedRuleUserList and may cause confusion. Part of the reason we ended up with the name AddUnionUserList is that I wrote several UserList-related examples all at the same time for a guide, and we wanted to make sure that all of the names were different enough as not to cause confusion.

PierrickVoulet: What about "AddLogicalUserList" instead?

devchas: Okay with me.

PierrickVoulet: @fiboknacky If OK with you, I will change to AddLogicalUserList as it seems to be the best tradeoff.

PierrickVoulet:
> Have you changed the example name? I've not seen it's changed yet.

Done, it should be good this time.

PR #392, opened 2020-07-29
Update AddCombinedRuleUserList.php

wfansh:
fix typo.


fiboknacky:
PBAND is already correct based on the code.
AND is a reserved word in PHP, and thus cannot be used.

PR #404, opened 2020-08-19
Install protobuf 3.11.4 PHP extension

PierrickVoulet:
This is avoid installing latest which is not adequate with the current version of the library: it has to be "~3.11.4".

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ env:
 before_script:
   - composer install -o -vvv --no-interaction
   - pecl install grpc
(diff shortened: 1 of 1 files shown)


fiboknacky:
Could you please track to change this back for the future releases?

fiboknacky (approved):

PierrickVoulet:
> Could you please track to change this back for the future releases?

Good point, I assumed that it would be caught during sync before releasing but I am also creating an issue to make sure we do not forget.

PR #411, opened 2020-08-31
Changes for release v5_0.

PierrickVoulet:


.travis.yml
```diff
@@ -20,7 +20,7 @@ env:
 before_script:
   - composer install -o -vvv --no-interaction
   - pecl install grpc
-  - pecl install protobuf-3.11.4
+  - pecl install protobuf
```
jradcliff, line 23: Should we specify `3.13.0` here, or is the issue that `pecl install` doesn't support composer-style version specifiers, so we have to decide what's best between a) and b) below?

a) no version here, with risk that we could pick up a breaking change

b) specifying a version here with the problem that we won't pick up improvements

PierrickVoulet: You are correct, `pecl install` does not support composer-style version.

We prefer `a)` for our Travis job because we want to catch any incompatibilities of the newest grpc and protobuf versions as soon as possible. Alternatively, `b)` can be used to workaround a known issue until it is fixed.

jradcliff (approved):

AnashOommen (approved):

PR #427, opened 2020-09-30
Added test cases for presence fields and modify FieldMasks to handle them

fiboknacky:
- Based on https://github.com/googleads/google-ads-java/pull/345/files and https://github.com/googleads/google-ads-java/blob/master/google-ads/src/test/resources/testdata/test_cases.textproto
- Files src/Google/Ads/GoogleAds/Util/Testing/Google/Ads/GoogleAds/Util/Testing/*.php are automatically generated by protoc.


src/Google/Ads/GoogleAds/Util/Testing/testdata/test_cases.json
```diff
@@ -108,5 +108,46 @@
       }
     },
     "expectedMask": "foo.bar.nestedWrapper"
+  }, {
```
PierrickVoulet, line 111: Could you add tests when things do not change?

I am thinking:

- Both set with same value
- Both unset

src/Google/Ads/GoogleAds/Util/Testing/Google/Ads/GoogleAds/Util/Testing/TestCase.php
```diff
@@ -75,7 +75,17 @@ public function setDescription($var)
      */
     public function getOriginalResource()
     {
-        return $this->original_resource;
```
PierrickVoulet, line 78: This does not seem to be optional in the proto, should not the code stay the same as before?

I noticed the same thing happens with other fields in TestCase, Resource, Foo and Bar.

src/Google/Ads/GoogleAds/Util/FieldMasks.php
```diff
@@ -247,4 +254,14 @@ private static function isWrapperType(Descriptor $descriptor)
     {
         return in_array($descriptor->getClass(), self::$WRAPPER_TYPES);
     }
+
+    // TODO: We can remove this function when it's supported in google/gax-php.
```
PierrickVoulet, line 258: Any gax-php issue # you point to? Could you create a new issue in the project so we can keep track of this?

src/Google/Ads/GoogleAds/Util/FieldMasks.php
```diff
@@ -155,9 +155,16 @@ private static function buildPaths(
                     $paths[] = $fieldName;
                 }
             } else {
+                $hasser = self::getHasser($fieldDescriptor->getName());
```
PierrickVoulet, line 158: This assumes that all fields use presence. I am not sure that we want to make this assumption?

PierrickVoulet (changes requested):

src/Google/Ads/GoogleAds/Util/Testing/Google/Ads/GoogleAds/Util/Testing/TestCase.php
fiboknacky: This is what the protoc generates. If you have concerns, let's discuss with the protobuf team.
My two-cents are that it's not quite different in PHP.
If the variable is unset and you get the variable as is, it's returned as null.

src/Google/Ads/GoogleAds/Util/Testing/testdata/test_cases.json
fiboknacky: With the test cases as JSON, 
* Both unset (I guess you mean the presence field) would be similar to when both original and modified resources are empty. We cannot set the presence field to null, as it requires to have an int type.
* Both set with same value --> This is now added.

src/Google/Ads/GoogleAds/Util/FieldMasks.php
fiboknacky: With the current protoc, it generates hasser for all fields anyway. Let's discuss this offline.

src/Google/Ads/GoogleAds/Util/FieldMasks.php
fiboknacky: From what you pointed out, let's discuss offline to find the conclusion before getting back to this again.

src/Google/Ads/GoogleAds/Util/FieldMasks.php
PierrickVoulet: Sure.

Quick clarification: this assumption could lead to issues if we do not regen all protos including older versions of the API because FieldMasks is not API version-specific.

src/Google/Ads/GoogleAds/Util/Testing/testdata/test_cases.json
PierrickVoulet: LGTM

src/Google/Ads/GoogleAds/Util/FieldMasks.php
fiboknacky: Thanks for reminding. I actually planned to add a condition to check if the method really exists too. Done now.

src/Google/Ads/GoogleAds/Util/Testing/Google/Ads/GoogleAds/Util/Testing/TestCase.php
PierrickVoulet: LGTM

PierrickVoulet (approved):

src/Google/Ads/GoogleAds/Util/FieldMasks.php
```diff
@@ -247,4 +255,15 @@ private static function isWrapperType(Descriptor $descriptor)
...
+
+    // TODO: We can remove this function when it's supported in google/gax-php:
+    // https://github.com/googleapis/gax-php/issues/285
+    /**
+     * @param string $name
+     * @return string the name of hasser function
+     */
+    private static function getHasser(string $name)
+    {
+        return 'has' . ucfirst(Serializer::toCamelCase($name));
```
jradcliff, line 267: Is this going to cover all cases? I remember there were some issues around PHP and method names, especially those that included a number.

fiboknacky: I'm not sure what cases you mean?
Anyway, this is the same as the methods implemented by gax-php (the only diff is that "has" is used instead of "get"), so if there are bugs in gax-php's `toCamelCase()`, we should prioritize fixing them in there.

But so far, after many releases, I've not found any bugs in that `toCamelCase()` method, so I'm convinced that this should be fine.

jradcliff: I was thinking of fields like `metrics.video_quartile_100_rate`, which we renamed to `metrics.video_quartile_p100_rate` (added `p` before `100`) for v5 to avoid ambiguity when converting to/from camel case and snake case.
Sounds like this isn't a concern, so I'll mark it resolved.

jradcliff (approved):

PR #433, opened 2020-10-05
Add UseCustomClientTimeouts code example

PierrickVoulet:
This new golden will be used to include code in the guides about timeouts e.g. [PHP](https://developers.google.com/google-ads/api/docs/client-libs/php/timeout).

Change-Id: I07b892b5d9e9286d040870b813837de82003a80f

### DIFF
--- a/examples/Misc/SetCustomClientTimeouts.php
+++ b/examples/Misc/SetCustomClientTimeouts.php
@@ -186,12 +186,15 @@ class SetCustomClientTimeouts
                     // When making an unary call, an optional argument is provided and can be
                     // used to override the default retry settings with given values.
                     'retrySettings' => [
-                        // Sets the timeout that is used for the first try.
-                        'initialRpcTimeoutMillis' => self::CLIENT_TIMEOUT_MILLIS,
-                        // Sets the maximum timeout that can be used for any given try.
-                        'maxRpcTimeoutMillis' => self::CLIENT_TIMEOUT_MILLIS,
                         // Sets the maximum accumulative timeout of the call, it includes all tries.
-                        'totalTimeoutMillis' => self::CLIENT_TIMEOUT_MILLIS
+                        'totalTimeoutMillis' => self::CLIENT_TIMEOUT_MILLIS,
+                        // Sets the timeout that is used for the first try to one tenth of the
+                        // maximum accumulative timeout of the call.
+                        'initialRpcTimeoutMillis' => self::CLIENT_TIMEOUT_MILLIS / 10,
+                        // Sets the maximum timeout that can be used for any given try to one fifth
+                        // of the maximum accumulative timeout of the call (two times greater than
+                        // the timeout that is used for the first try).
+                        'maxRpcTimeoutMillis' => self::CLIENT_TIMEOUT_MILLIS / 5
                     ]
                 ]
             );
--- a/examples/Misc/SetCustomClientTimeouts.php
+++ b/examples/Misc/SetCustomClientTimeouts.php
@@ -33,12 +33,18 @@ use Google\Ads\GoogleAds\V5\Services\GoogleAdsRow;
 use Google\ApiCore\ApiException;
 use Google\ApiCore\ApiStatus;
 
-/** This example illustrates the use of custom client timeouts. */
-class UseCustomClientTimeouts
+/**
+ * This example illustrates the use of custom client timeouts in the context of server streaming and
+ * unary calls.
+ *
+ * For more information about the concepts, see this documentation:
+ * https://grpc.io/docs/what-is-grpc/core-concepts/#rpc-life-cycle.
+ */
+class SetCustomClientTimeouts
 {
     private const CUSTOMER_ID = 'INSERT_CUSTOMER_ID_HERE';
 
-    private const CUSTOM_CLIENT_TIMEOUT = 5 * 60 * 1000; // 5 minutes in millis
+    private const CLIENT_TIMEOUT_MILLIS = 5 * 60 * 1000; // 5 minutes in millis
 
     public static function main()
     {
@@ -98,8 +104,8 @@ class UseCustomClientTimeouts
      */
     public static function runExample(GoogleAdsClient $googleAdsClient, int $customerId)
     {
-        self::serverStreamingCall($googleAdsClient, $customerId);
-        self::unaryCall($googleAdsClient, $customerId);
+        self::makeServerStreamingCall($googleAdsClient, $customerId);
+        self::makeUnaryCall($googleAdsClient, $customerId);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


examples/Misc/UseCustomClientTimeouts.php
```diff
@@ -0,0 +1,190 @@
...
+use Google\Ads\GoogleAds\Lib\V5\GoogleAdsException;
+use Google\Ads\GoogleAds\Lib\OAuth2TokenBuilder;
+use Google\Ads\GoogleAds\Lib\V5\GoogleAdsServerStreamDecorator;
+use Google\Ads\GoogleAds\V5\Errors\GoogleAdsError;
+use Google\Ads\GoogleAds\V5\Services\GoogleAdsRow;
+use Google\ApiCore\ApiException;
+use Google\ApiCore\ApiStatus;
+
+/** This example illustrates the use of custom client timeouts. */
+class UseCustomClientTimeouts
```
fiboknacky, line 37: SetCustomClientTimeouts seems a bit better to me.

examples/Misc/UseCustomClientTimeouts.php
```diff
@@ -0,0 +1,190 @@
...
+{
+    private const CUSTOMER_ID = 'INSERT_CUSTOMER_ID_HERE';
+
+    private const CUSTOM_CLIENT_TIMEOUT = 5 * 60 * 1000; // 5 minutes in millis
```
fiboknacky, line 41: Can this be CLIENT_TIMEOUT_MILLIS?
In this way, it's clear from the name that this is millis.

examples/Misc/UseCustomClientTimeouts.php
```diff
@@ -0,0 +1,190 @@
...
+
+    /**
+     * Runs the example.
+     *
+     * @param GoogleAdsClient $googleAdsClient the Google Ads API client
+     * @param int $customerId the customer ID
+     */
+    public static function runExample(GoogleAdsClient $googleAdsClient, int $customerId)
+    {
+        self::serverStreamingCall($googleAdsClient, $customerId);
```
fiboknacky, line 101: The method name should start with a verb. Could you make it "makeServerStreamingCall"? (Same for below).

examples/Misc/UseCustomClientTimeouts.php
```diff
@@ -0,0 +1,190 @@
...
+use Google\Ads\GoogleAds\Lib\V5\GoogleAdsClientBuilder;
+use Google\Ads\GoogleAds\Lib\V5\GoogleAdsException;
+use Google\Ads\GoogleAds\Lib\OAuth2TokenBuilder;
+use Google\Ads\GoogleAds\Lib\V5\GoogleAdsServerStreamDecorator;
+use Google\Ads\GoogleAds\V5\Errors\GoogleAdsError;
+use Google\Ads\GoogleAds\V5\Services\GoogleAdsRow;
+use Google\ApiCore\ApiException;
+use Google\ApiCore\ApiStatus;
+
+/** This example illustrates the use of custom client timeouts. */
```
fiboknacky, line 36: Can you provide a bit more description about the types of calls of which we show how to set the client timeouts.

examples/Misc/UseCustomClientTimeouts.php
```diff
@@ -0,0 +1,190 @@
...
+     */
+    private static function serverStreamingCall(GoogleAdsClient $googleAdsClient, int $customerId)
+    {
+        $googleAdsServiceClient = $googleAdsClient->getGoogleAdsServiceClient();
+        // Creates a query that retrieves all campaign IDs.
+        $query = 'SELECT campaign.id FROM campaign';
+
+        $output = '';
+        try {
+            // Issues a search stream request.
```
fiboknacky, line 119: Issues a search stream request by setting a custom client timeout.

examples/Misc/UseCustomClientTimeouts.php
```diff
@@ -0,0 +1,190 @@
...
+     */
+    private static function unaryCall(GoogleAdsClient $googleAdsClient, int $customerId)
+    {
+        $googleAdsServiceClient = $googleAdsClient->getGoogleAdsServiceClient();
+        // Creates a query that retrieves all campaign IDs.
+        $query = 'SELECT campaign.id FROM campaign';
+
+        $output = '';
+        try {
+            // Issues a search request.
```
fiboknacky, line 158: Ditto. There are some settings that are different from normal code examples.
We probably can explain a bit more here.

examples/Misc/UseCustomClientTimeouts.php
PierrickVoulet: Definitely better, changing it.

examples/Misc/SetCustomClientTimeouts.php
```diff
@@ -147,24 +163,35 @@ private static function serverStreamingCall(GoogleAdsClient $googleAdsClient, in
...
+                    // used to override the default retry settings with given values.
                     'retrySettings' => [
-                        'initialRpcTimeoutMillis' => self::CUSTOM_CLIENT_TIMEOUT,
-                        'maxRpcTimeoutMillis' => self::CUSTOM_CLIENT_TIMEOUT,
-                        'totalTimeoutMillis' => self::CUSTOM_CLIENT_TIMEOUT
+                        // Sets the timeout that is used for the first try.
+                        'initialRpcTimeoutMillis' => self::CLIENT_TIMEOUT_MILLIS,
+                        // Sets the maximum timeout that can be used for any given try.
+                        'maxRpcTimeoutMillis' => self::CLIENT_TIMEOUT_MILLIS,
+                        // Sets the maximum accumulative timeout of the call, it includes all tries.
```
fiboknacky, line 193: If I guess correctly, once the call fails, it will use the initialRpcTimeoutMillis option to calculate the next timeout to wait, which is supposed to be larger than initialRpcTimeoutMillis, right?

If so, would it make more sense to set the three options differently?

PierrickVoulet: There is no one-fit-all configuration. Here, I use the same as what we generate by default but with a custom value (see the `3600000` occurences in this [example](https://github.com/googleads/google-ads-php/blob/master/src/Google/Ads/GoogleAds/V2/Services/resources/ad_service_client_config.json#L16)). We set `initialRpcTimeoutMillis` and `maxRpcTimeoutMillis` to the maximum timeout because we want to maximize the timeout used by the tries from the really first try, this is to maximize our chances of success. If a try fails and there is some time left, the next try will use a timeout that will be lower than `initialRpcTimeoutMillis` because capped by the accumulative timeout `maxRpcTimeoutMillis`.

Another option would be to disable retries overriding the setting `retriesEnabled` for the call. This would be less confusing because only one setting `timeoutMillis` would then be used like for `GAS.SearchStream`. But it will be less resilient because it will not retry if a transient error occurs during the call (unrelated to timeout).

What do you think?

fiboknacky: I didn't realize that the three params were also set to be the same by default.
I'm fine with it, although I just think that if we can at least separate `maxRpcTimeoutMillis` from `total_timeout_millis`, it will provide us venue to explain the relationships between them for beginners.

As an super simplified example, if `maxRpcTimeoutMillis` is 1 and `total_timeout_millis` is 5, you can explain that in general, this will retry around 5 times until it reaches 5 millis, which is `total_timeout_millis`.

PierrickVoulet: Sounds good, I edited to reflect that.

fiboknacky (approved):

AnashOommen (approved):

PR #456, opened 2020-11-16
Add GetInvoices code example

PierrickVoulet:


examples/Billing/GetInvoices.php
```diff
@@ -0,0 +1,212 @@
...
+        // [START GetInvoices]
+        // Issues the request.
+        $response = $googleAdsClient->getInvoiceServiceClient()->listInvoices(
+            $customerId,
+            ResourceNames::forBillingSetup($customerId, $billingSetupId),
+            // The year needs to be 2019 or later.
+            date('Y', $lastMonth),
+            MonthOfYear::value(strtoupper(date('F', $lastMonth)))
+        );
+        // [END GetInvoices]
```
fiboknacky, line 122: This is not exactly related to the example, but WDYT about having the explanation of this kind of region tags somewhere in our DevSite?
This would be a bit more friendly for any code example users.

examples/Billing/GetInvoices.php
```diff
@@ -0,0 +1,212 @@
...
+
+        // [START GetInvoices_1]
+        // Iterates over all invoices retrieved and prints their information.
+        foreach ($response->getInvoices() as $invoice) {
+            /** @var Invoice $invoice */
+            printf(
+                "- Found the invoice '%s':" . PHP_EOL .
+                "  ID (also known as Invoice Number): %s" . PHP_EOL .
```
fiboknacky, line 130: Is it a string like primary billing ID that has hyphens?
If so, could you add a single quote?
Please do the same for other IDs in this example.

examples/Billing/GetInvoices.php
```diff
@@ -0,0 +1,212 @@
...
+                    self::microsToPlain($accountBudgetSummary->getTaxAmountMicros()),
+                    self::microsToPlain($accountBudgetSummary->getTotalAmountMicros())
+                );
+            }
+        }
+        // [END GetInvoices_1]
+    }
+
+    /**
+     * Converts an amount from the micro unit to the plain unit.
```
fiboknacky, line 201: Nit: I don't quite get the word "plain" at first.
Do we have a bit verbose explanation? Like "the unit of your currency set in your account preferences"?

examples/Billing/GetInvoices.php
PierrickVoulet: I like the idea. I could add a note on this [page](https://developers.google.com/google-ads/api/docs/samples) for reference, what do you think?

examples/Billing/GetInvoices.php
PierrickVoulet: The technical term seems to be "base" instead fo "plain", does that help if I change?

examples/Billing/GetInvoices.php
fiboknacky: SGTM.

examples/Billing/GetInvoices.php
fiboknacky: Yep. That helps.

examples/Billing/GetInvoices.php
PierrickVoulet: Done. I also did it for the fields that may have spaces in their values (e.g., names, descriptions).

fiboknacky (approved):

examples/Billing/GetInvoices.php
PierrickVoulet: Done, let's continue this discussion offline.

sundquist (approved):
LGTM

PR #466, opened 2020-11-20
Add print_php_information script

PierrickVoulet:
The purpose of this script is to provide the means for our users to easily gather information about their PHP environment and `protobuf` / `grpc` extensions. This would typically be used at issue creation time.

### DIFF
--- a/scripts/print_php_information.php
+++ b/scripts/print_php_information.php
@@ -33,10 +33,10 @@ function printExtensionInformation(string $extensionName)
 }
 
 // Prints the general information about PHP.
-print '================= PHP GENERAL INFORMATION =====' . PHP_EOL;
+print '================= PHP GENERAL INFORMATION' . PHP_EOL;
 phpinfo(INFO_GENERAL) . PHP_EOL;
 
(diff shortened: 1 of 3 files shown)


scripts/print_php_information.php
```diff
@@ -0,0 +1,42 @@
...
+        'The PHP Extension %s is %sinstalled%s%s',
+        $extensionName,
+        $isInstalled ? '' : 'not ',
+        $isInstalled ? ': ' . phpversion($extensionName) : '',
+        PHP_EOL
+    );
+}
+
+// Prints the general information about PHP.
+print '================= PHP GENERAL INFORMATION =====' . PHP_EOL;
```
fiboknacky, line 36: Nit: Is there a reason why the left "====" is longer than the right "==="?

PierrickVoulet: Not really. What about I only keep the one on the left side to simplify?

fiboknacky: That's fine too. I just feel it's a bit strange to have different numbers of equal signs. :)

fiboknacky (approved):

PR #472, opened 2020-11-24
Add region tags

PierrickVoulet:


jradcliff (approved):

examples/AdvancedOperations/AddSmartDisplayAd.php
```diff
@@ -258,6 +262,7 @@ private static function createSmartDisplayCampaign(
      * @param string $campaignResourceName the resource name of the campaign
      * @return string the resource name of the newly created ad group
      */
+    // [START AddSmartDisplayAd_4]
```
fiboknacky, line 265: Just curious, why 2 and 3 happen below 4 ?

fiboknacky (approved):

examples/AdvancedOperations/AddSmartDisplayAd.php
PierrickVoulet: The migration script generates the region tag names as they are discovered when crawling the DevSite.

PR #483, opened 2020-12-03
Upgrade FieldMask Unit Tests and fix false boolean comparison

### LINKED ISSUE #480: FieldMasks::allSetFieldsOf not setting enhancedCpcEnabled > false
When using the `FieldMasks::allSetFieldsOf` to set a `ManualCpc` on a campaign, it will not update the campaign when `enhancedCpcEnabled` is `false`. However, if this field is `true`, it does. See code below to reproduce. I would expect that if you turn Enhanced CPC off, it would still see it as an update.

**What versions of the client library and Google Ads API are you using?**

v5.0.0 and V5
I checked the `FieldMasks` of V6 and it does the same thing

**What environment are you using?**

- PHP 7.4.12
- Linux 3.10.0-1062.18.1.el7.x86_64 #1 SMP Tue Mar 17 23:49:17 UTC 2020 x86_64
- Protobuf; C version 3.14.0
- gRPC version 1.33.1

**Actions taken**

```
$googleCampaign = new Campaign();
$googleCampaign->setResourceName(ResourceNames::forCampaign($accountId, $campaignId));

$manualCpc = new ManualCpc();
$manualCpc->setEnhancedCpcEnabled($biddingStrategyModel->isEnhancedCpc());

$googleCampaign->setManualCpc($manualCpc);

var_dump(FieldMasks::allSetFieldsOf($googleCampaign)->serializeToJsonString());
```

**Expected result**
```
"resourceName,manualCpc.enhancedCpcEnabled"
```

**Actual result**
```
"resourceName"
```

PierrickVoulet:
This fixes multiple issues at once:
- #480 : It is only possible to check the difference between an unset and a false boolean by using the hasser. Non optional-booleans are infeasible because the default value of a bool is false in Protobuf but this is not an issue because the Google Ads API uses proto3 presence everywhere since V6.
- Test cases were using the library field mask comparison utils to generate the expected results at runtime: This leaded to much confusion when something was wrong so I added a new field instead to store hard-coded masks for all test cases.
- One assert generated false positive results when comparing field mask objects: I switched to a string comparison of the field masks instead.
- Some type hints were missing: I added them.

Change-Id: I64f83d507e1cf6cf802d0b607c7b78900f443334


src/Google/Ads/GoogleAds/Util/FieldMasks.php
```diff
@@ -163,6 +163,7 @@ private static function buildPaths(
                     // beginning.
                     (!is_null($original) && method_exists($original, $hasser)
                         && $original->$hasser() != $modified->$hasser())
+                    || (is_null($original) && !$modified)
```
fiboknacky, line 166: Why do we check for `$modified` here?
Why not `$modifiedValue`?

I ask because above, we also obtained the `$modifiedValue`, which is the result of `$modified->$getter()`, so probably we don't expect this `$modified` to be a primitive value.

PierrickVoulet: You are right, I changed.

src/Google/Ads/GoogleAds/Util/FieldMasks.php
```diff
@@ -163,6 +163,7 @@ private static function buildPaths(
                     // beginning.
                     (!is_null($original) && method_exists($original, $hasser)
                         && $original->$hasser() != $modified->$hasser())
+                    || (is_null($original) && !$modifiedValue)
```
fiboknacky, line 166: I'm a bit confused by this.
What are we trying to check here?
Looks like if original is null and the modified value is false then this will be marked as a value changed.
What about the case when the modified value is true?

PierrickVoulet: This is to alleviate the fact that when the original is non-existent and the modified value is  a `false` boolean then the `$originalValue != $modifiedValue` condition returns `false`.

This is due to some inconsistency in PHP's `==` operator:

```php
print (null == false);

//> 1
```

fiboknacky: Then, why don't we just modify the code below to

```php
|| $originalValue !== $modifiedValue;
```

instead?
That should work for both `true` and `false` cases, without the need to add more conditions.

fiboknacky: Yep, please do.

If necessary, we can split comparison when the field type is MESSAGE and when it's not, for clearer logic.

PierrickVoulet: The confusion came from how the unit tests were written: the expected values were computed at runtime by the library instead of hard-coded and used an assert that was always returning `true` even if untrue. I upgraded the tests to avoid these issues.

We now have one clear test case that reproduces the exact problem with both implementations of protobuf.

src/Google/Ads/GoogleAds/Util/FieldMasks.php
```diff
@@ -159,10 +159,15 @@ private static function buildPaths(
                 $hasser = self::getHasser($fieldDescriptor->getName());
                 $hasValueChanged =
                     // In the first condition, when $original is null, $modified will not be null.
+                    // In the second condition, when $origin is null, we just need to check if
```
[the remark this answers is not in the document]
fiboknacky: And I think the existing comment is not quite updated. It talks about the case when $original is null, but the condition seems to be for the case when $origin is NOT null.
Could you update that comment too?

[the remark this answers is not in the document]
fiboknacky: And "the deep comparison done" means

```
originalValue != $modifiedValue
```

?

Just want to make sure we understand the same thing.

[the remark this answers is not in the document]
PierrickVoulet: I updated the comments to clarify. Yes, by deep comparison I mean the `!=` operator.

fiboknacky (approved):

PR #488, opened 2020-12-07
Remove SHOULD_ prefix of argument SHOULD_DELETE_EXISTING_FEEDS

PierrickVoulet:
Change-Id: I21f2d5226feaa5d0d0e50772e10645f38aedbe1f


fiboknacky (changes requested):
You probably need to change ArgumentNames.php too?

PierrickVoulet:
> You probably need to change ArgumentNames.php too?

Good catch, I forgot to add this file before committing! Done.

fiboknacky (approved):

PR #489, opened 2020-12-07
Fix #353

### LINKED ISSUE #353: Composer Deprecation Warnings for PSR-4 Namespace
Composer reports that a few files in the Testing directory have namespaces that aren't PSR-4 compliant. This can be solved from the user side in the short term in various ways, but the long term fix will need to be making the namespaces PSR-4 compliant.
```
Class Google\Ads\GoogleAds\Util\Testing\TestSuite ./vendor/googleads/google-ads-php/src/Google/Ads/GoogleAds/Util/Testing/Google/Ads/GoogleAds/Util/Testing/TestSuite.php
Class Google\Ads\GoogleAds\Util\Testing\Resource ./vendor/googleads/google-ads-php/src/Google/Ads/GoogleAds/Util/Testing/Google/Ads/GoogleAds/Util/Testing/Resource.php
Class Google\Ads\GoogleAds\Util\Testing\TestCase ./vendor/googleads/google-ads-php/src/Google/Ads/GoogleAds/Util/Testing/Google/Ads/GoogleAds/Util/Testing/TestCase.php
Class Google\Ads\GoogleAds\Util\Testing\Foo ./vendor/googleads/google-ads-php/src/Google/Ads/GoogleAds/Util/Testing/Google/Ads/GoogleAds/Util/Testing/Foo.php
Class Google\Ads\GoogleAds\Util\Testing\Bar ./vendor/googleads/google-ads-php/src/Google/Ads/GoogleAds/Util/Testing/Google/Ads/GoogleAds/Util/Testing/Bar.php
```

PierrickVoulet:
I chose to move the resources that are only used during development for testing purposes to the `tests` directory. This includes the generator script, README, test cases JSON, proto and generated code, etc. I did this because it makes everything cleaner and well scoped in the hierarchy and in the configurations for package management (composer) and code style.

To exclude those resources from the [classmap](https://getcomposer.org/doc/04-schema.md#exclude-files-from-classmaps) should fix the PSR-4 compliance warnings but I do not think that it is a real fix. I think that having all tests resources in `tests` makes it clean and easier to maintain, not just PSR-4 compliant.

Change-Id: I6662072cbabce622dd762c75628373ef44665d6a


fiboknacky:
Could you share your thoughts why you changed in this way?
In particular, do you intend to move _all_ testing files out of `src/`?

I have no objections of doing that but

1) We should have more descriptions about this change
2) Why we want to do this way (apart from fixing PSR-4 issues). Is changing file location the only way to fix this PSR-4 issues?
3) It looks like there are repos that put testing files in `src/` and some that put testing files in `tests/`. What justification that we use to change the file locations?

fiboknacky (commented):
Added some comments.

PierrickVoulet:
> Could you share your thoughts why you changed in this way?

I chose to move the resources that are only used during development for testing purposes to the `tests` directory. This includes the generator script, README, test cases JSON, proto and generated code. I did this because it makes everything cleaner and well scoped in the hierarchy and in the configurations for package management (composer) and code style.

> In particular, do you intend to move _all_ testing files out of `src/`?

Yes, did you find anything that I missed in this PR?

> I have no objections of doing that but
> 
> 1. We should have more descriptions about this change

Do you mean in this PR or somewhere else? If in this PR, I can add more details in the description after we have a consensus.

> 2. Why we want to do this way (apart from fixing PSR-4 issues). Is changing file location the only way to fix this PSR-4 issues?
> 3. It looks like there are repos that put testing files in `src/` and some that put testing files in `tests/`. What justification that we use to change the file locations?

To exclude those resources from the [classmap](https://getcomposer.org/doc/04-schema.md#exclude-files-from-classmaps) should fix the PSR-4 compliance warnings but I do not think that it is a real fix. I think that having all tests resources in `tests` makes it clean and easier to maintain, not just PSR-4 compliant.

fiboknacky:
> > In particular, do you intend to move _all_ testing files out of `src/`?
> 
> Yes, did you find anything that I missed in this PR?

Yes, what I meant by moving all testing files is that all files that are in `Testing` subdirectories (e.g., `Lib/Testing`), not just in `Util` that you're moving.
If you have some testing files in `tests/` but some are still `src/`, that will be inconsistent and can confuse users or contributors in the future.
I'd just move all files out of `src/` or not at all.
 
> Do you mean in this PR or somewhere else? If in this PR, I can add more details in the description after we have a consensus.

Yes, the justification like you explained above would be just fine. :) 

> To exclude those resources from the classmap should fix the PSR-4 compliance warnings but I do not think that it is a real fix. I think that having all tests resources in tests makes it clean and easier to maintain, not just PSR-4 compliant.

Yep. That's what I mean. I'm not rooting for either way purely, as I do see both ways are used by some well-known repos.
But if we're doing this, we should ensure we're consistent.

It's not needed to be in this PR, but could you own this and do the rest of the change later soon?

composer.json
```diff
@@ -32,8 +32,8 @@
   },
   "autoload-dev": {
     "psr-4": {
-      "Google\\": "src/Google/Ads/GoogleAds/Util/Testing/Google",
-      "GPBMetadata\\": "src/Google/Ads/GoogleAds/Util/Testing/metadata"
+      "Google\\Ads\\GoogleAds\\": "tests/Google/Ads/GoogleAds",
+      "GPBMetadata\\Google\\Ads\\GoogleAds\\": "tests/metadata/Google/Ads/GoogleAds"
```
fiboknacky, line 36: Just in case: This doesn't have bad effect with normal source and example files, right?

phpcs_ruleset.xml
```diff
@@ -3,8 +3,8 @@
     <rule ref="PSR12"/>
     <exclude-pattern>metadata</exclude-pattern>
     <exclude-pattern>src/Google/Ads/GoogleAds/V[0-9]+</exclude-pattern>
-    <exclude-pattern>src/Google/Ads/GoogleAds/Util/Testing/Google</exclude-pattern>
-    <exclude-pattern>src/Google/Ads/GoogleAds/Util/Testing/metadata</exclude-pattern>
+    <exclude-pattern>tests/metadata</exclude-pattern>
+    <exclude-pattern>tests/Google/Ads/GoogleAds/Util/FieldMasks/Proto</exclude-pattern>
```
fiboknacky, line 7: Do you need to account for the files you moved from Lib/Testing?

tests/Google/Ads/GoogleAds/Lib/ConfigurationLoaderTestProvider.php
```diff
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-namespace Google\Ads\GoogleAds\Lib\Testing;
+namespace Google\Ads\GoogleAds\Lib;
```
fiboknacky, line 19: I think we should still somehow separate these helping files from the rest of normal unit tests.
And then add a rule in phpcs_ruleset.xml to exclude them.
The folder name may not need to be `Testing`, but it should indicate it's the place for it's test helpers.

Currently, if a user runs phpunit without specifying the directories, these files will be unnecessarily included.
Although it would work, it includes too many files than it needs.

phpcs_ruleset.xml
PierrickVoulet: These exclusions only target the generate code. We still want to check the code style of those and I believe it is the case with the current rules.

tests/Google/Ads/GoogleAds/Lib/ConfigurationLoaderTestProvider.php
PierrickVoulet: Why should we exclude them from code style? I believe that those are not excluded today and are compliant.

I think this would add complexity for small-to-no benefits. Did you keep a link to a well designed implementation that does this? If  we go with this solution, I think of two options: add a new directory next to `tests` or have a `Google\Ads\GoogleAds\Testing` package so that all those classes are at the same location, any preference on your side?

composer.json
PierrickVoulet: Not that I could find, at least `GetCampaigns.php` runs fine. Any particular reason you think it could have an impact?

fiboknacky: They're the same as those rules in "psr-4" of "autoload", so I just wonder if they play along together well.
Previously, they're different rules.

phpcs_ruleset.xml
fiboknacky: Oops. I misread this as a phpunit file. Sorry.

tests/Google/Ads/GoogleAds/Lib/ConfigurationLoaderTestProvider.php
fiboknacky: Sorry. I don't mean the code style checking, but the PHPUnit testing.
With your structure, can it separate what are really unit tests and what are not?

fiboknacky: NVM. I think we can use something similar to [this line](https://github.com/sebastianbergmann/comparator/blob/master/phpunit.xml#L15) to filter those files.
Let me TAL at your PR one more time.

fiboknacky: All look fine. Could you modify phpunit rules to be more specific?

phpcs_ruleset.xml
PierrickVoulet: No problem, resolving this comment.

composer.json
PierrickVoulet: I think the best is to regroup all testing classes that are not actual tests under the package `Google\Ads\GoogleAds\Tests` as suggested by Composer `autoload-dev` documentation [here](https://getcomposer.org/doc/04-schema.md#autoload-dev). What do you think?

tests/Google/Ads/GoogleAds/Lib/ConfigurationLoaderTestProvider.php
PierrickVoulet: I refreshed the phpunit configuration file: I removed unnecessary coverage whitelist exclusions and I added the `Test.php` suffix filter as you suggested.

composer.json
fiboknacky: I don't think that doc recommends grouping all non-test into `Tests`, but grouping any helping files into one place would be fine, of course.
However, as I mentioned in the comment thread below, you probably don't need to do that, as we can specify a suffix in the rule.

So, let's go with this version you have.

fiboknacky (approved):

composer.json
PierrickVoulet: I will wait for the code coverage report to be updated following my last commit to be sure. If there is a drop I suggest that we regroup those helper classes in this single package and add a rule to whitelist exclude it from the phpunit coverage.

PierrickVoulet: The code coverage changed but it is unrelated to the change so I keep it like that for now. We can further improve this later on if needed.

PR #493, opened 2020-12-10
Fix 481 Introduce new interceptor for lazy init of GGAFailures

### LINKED ISSUE #481: Lazily init GoogleAdsFailure from gRPC interceptor
**Describe the problem you are trying to solve**

As originally found in #469 the descriptor pool is lazily initialized with the class `GoogleAdsFailure` at log level `DEBUG` but not at coarser log levels (e.g., `INFO`, `ERROR`): this can lead to unexpected behaviors.

**Describe the solution you would like**

We need to start by identifying all the locations where the class `GoogleAdsFailure` is initialized in the descriptor pool and make sure that it covers all the use cases.

Typically, changing the initialization done in the logger to a gRPC interceptor seems to be a good overall improvement.

PierrickVoulet:
Fixes #481 

This expands the lazy init that was done when logging the details of responses to all unary calls (not needed by streaming calls).

Change-Id: I848af6978c6e31f4410aac6043c4c7c5d570f2fb


PierrickVoulet:
@fiboknacky I only did the job for V6 and did not include any unit tests for now. I will do it when we meet consensus.

fiboknacky (approved):
LGTM.

PierrickVoulet:
@fiboknacky I expanded the change to all versions and added unit tests. Feel free to review and merge if OK with you.

fiboknacky (approved):

PR #501, opened 2020-12-16
Update Helper.php

devorel:
Can you add it ?

### DIFF
--- a/examples/Utils/Helper.php
+++ b/examples/Utils/Helper.php
@@ -42,4 +42,12 @@ final class Helper
     {
         return (new DateTime())->format("mdHisv");
     }
+    
+        public static function toMicro($number) {//min 10000 //1000000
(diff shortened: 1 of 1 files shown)


PierrickVoulet:
Hi @devorel - Thank you for the suggestion and writing this implementation. Quick question to better understand your intent: do you need these helpers for the development of your own applications?

devorel:
hi yes
I use with it , it's prevents errors

PierrickVoulet:
Thanks for the clarification and feedback, @devorel . In fact, something similar was developed for the code example [GetInvoices](https://github.com/googleads/google-ads-php/blob/master/examples/Billing/GetInvoices.php#L206) so we will plan on adding those soon. We will you keep posted.

PierrickVoulet:
I opened another PR #506 following your lead, I suggest we continue working on this improvement from there. Thanks a lot for the contribution!

PR #506, opened 2020-12-21
Added micro converter helpers for code examples

PierrickVoulet:
Initially suggested and implemented from #501 (thanks for contributing @devorel!)

This PR replaces #501:
- Reuses the logic of conversion initially implemented for the `GetInvoices` code example alone
- Refactors all the code examples that explicitly uses the micro to base or base to micro conversions


examples/Utils/Helper.php
```diff
@@ -42,4 +42,26 @@ public static function getShortPrintableDatetime(): string
     {
         return (new DateTime())->format("mdHisv");
     }
+
```
devorel, line 45: ### 2 bugs!

try the function microToBase with the number **2246197**
its return **2**, but i need get **2.25** (GoogleAPI works with 2 decimals)


my function do it well!
```
    public static function microToBase($number) {
        return (float) number_format($number / 1000000, 2, '.', '');
    }
```
try the function baseToMicro with the number **0.1**
its return **0**, but i need get **100000** (GoogleAPI works with 2 decimals- [if you try more ,you get error] ) 


my function do it well!
```
    public static function baseToMicro($number) {//min 10000
        return (int) str_replace('.', '', number_format($number, 2, '.', '') . '0000');
    }
```
i understand my code it's not alginitic , but it works well

PierrickVoulet: Thanks for checking, I am adding unit tests and will update the logic accordingly.

PierrickVoulet: Google Ads API uses micro unit integers, not floating decimal. I upgraded the methods to:
- Take both float and int amounts as input
- Return float with `microToBase`

This should fix both issues.

tests/Google/Ads/GoogleAds/Examples/Utils/HelperTest.php
```diff
@@ -0,0 +1,69 @@
...
+     */
+    public function testMicroToBase($amount, float $expectedResult)
+    {
+        $this->assertEquals($expectedResult, Helper::MicroToBase($amount));
+    }
+
+    public function microToBaseCases()
+    {
+        return [
+            [123456, 0.123456],
```
fiboknacky, line 44: Add a case of 0 and null to both test functions too.

fiboknacky (changes requested):

fiboknacky (approved):

Issue #512, opened 2021-01-08
getting Fatal error while running getChangeDetails for change events


Vijaysinh:
![error](https://user-images.githubusercontent.com/6183730/104047750-2b069600-5208-11eb-8583-1359b0ea8a7d.PNG)
I am getting api response but same time after printing API response I am getting below error.

I am using PHP.

I am using V6 api version.

Fatal error: Uncaught TypeError: Argument 1 passed to Google\Ads\GoogleAds\Util\FieldMasks::getDescriptorForMessage() must be an instance of Google\Protobuf\Internal\Message, instance of Google\Protobuf\Internal\RepeatedField given, called in C:\wamp64\www\libs\google-ads-php\src\Google\Ads\GoogleAds\Util\FieldMasks.php on line 261 and defined in C:\wamp64\www\libs\google-ads-php\src\Google\Ads\GoogleAds\Util\FieldMasks.php on line 293

fiboknacky:
Do you use the [C implementation of protobuf](https://developers.google.com/google-ads/api/docs/client-libs/php/protobuf#c_implementation)? And what PHP version do you use?

Vijaysinh:
@fiboknacky I am using php 7.3.21 and I am using  PHP implementation using composer.

This is my composer.json require json,

"require": {
    "php": ">=7.2",
    "google/gax": "^1.5.0",
    "google/protobuf": "^3.14.0",
    "ulrichsg/getopt-php": "^3.4",
    "monolog/monolog": "^1.23.0 || ^2.0"
  },

fiboknacky:
BTW, what [library version](https://github.com/googleads/google-ads-php/releases) do you use and could you tell your customer ID that you use to run this example?

Vijaysinh:
@fiboknacky I am using v6 version of ads api and customerID = 1255040287

fiboknacky:
Thanks for the information.
For the version, I meant the [library version](https://github.com/googleads/google-ads-php/releases), not the Google Ads API version. Could you follow the link in the previous sentence and confirm what the library version you downloaded?

Vijaysinh:
I am using google-ads-php-v6.1.0 library version

fiboknacky:
We're looking into fixing this soon. Sorry for your inconvenience.
For now, you can just replace [this line](https://github.com/googleads/google-ads-php/blob/master/src/Google/Ads/GoogleAds/Util/FieldMasks.php#L255), with `return null`.
[...]

PR #519, opened 2021-01-26
Fixed the FieldMasks util's getFieldValue 

fiboknacky:
To make it work with null nested message and a repeated field of nested message.

### DIFF
--- a/tests/Google/Ads/GoogleAds/Util/FieldMasksTest.php
+++ b/tests/Google/Ads/GoogleAds/Util/FieldMasksTest.php
@@ -21,7 +21,6 @@ namespace Google\Ads\GoogleAds\Util;
 use Google\Ads\GoogleAds\Util\FieldMasks\FieldMasksTestDataProvider;
 use Google\Ads\GoogleAds\Util\FieldMasks\Proto\Resource;
 use Google\Ads\GoogleAds\Util\FieldMasks\Proto\TestSuite;
-use Google\Ads\GoogleAds\V6\Common\CustomParameter;
 use Google\Ads\GoogleAds\V6\Common\ExpandedTextAdInfo;
 use Google\Ads\GoogleAds\V6\Common\PercentCpc;
 use Google\Ads\GoogleAds\V6\Enums\AdvertisingChannelTypeEnum\AdvertisingChannelType;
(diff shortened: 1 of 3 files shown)


tests/Google/Ads/GoogleAds/Util/FieldMasksTest.php
```diff
@@ -135,15 +137,32 @@ public function testGetFieldValue()
                     ->getIterator()
             )
         );
+        // A value of a repeated field type of a message can be obtained.
```
PierrickVoulet, line 140: Could you also add a unit test that covers when the field value  path traverses an array field?

Typically, the field value paths `url_custom_parameters.key` and `url_custom_parameters[0].key` should raise the `UnexpectedValueException` exception.

PierrickVoulet (changes requested):

tests/Google/Ads/GoogleAds/Util/FieldMasksTest.php
fiboknacky: `url_custom_parameters[0].key` should raise the exception because `xxx[0]` is originally not the form that this method can handle. I think we need to first confirm if this form is ever returned by the API.

This fix aims to at least able to handle the repeated field of messages. I think we need to first define if `xxx[0]` is a valid field mask that can be returned by the ChangeEvent resource.
I tried to reproduce #512, but unfortunately I cannot get the real use case of it too, so I'm not sure that the user gets a repeated field of null (messages) or if the user really gets a repeated field of non-empty messages (with a path like `url_custom_parameters[0].key`, as you mentioned.

@AnashOommen WDYT? Looks like you're the one who worked on this example last time?

AnashOommen (approved):

tests/Google/Ads/GoogleAds/Util/FieldMasksTest.php
fiboknacky: Syncing with Anash offline. Turns out Anash believes the repeated fields shouldn't be returned as a field mask path by ChangeEvent. So I'm gonna roll the change about it back.
We can get back to this one we have more evidence that the field mask path can be repeated fields.

PierrickVoulet (approved):

PR #552, opened 2021-03-08
Rearrange methods of resource names utility in alphabetic order and add missing unit tests

PierrickVoulet:
Change-Id: I8018f8f66d34bb885064609f05f224230cdf167d

### DIFF
--- a/tests/Google/Ads/GoogleAds/Util/V6/ResourceNamesTest.php
+++ b/tests/Google/Ads/GoogleAds/Util/V6/ResourceNamesTest.php
@@ -20,6 +20,7 @@ namespace Google\Ads\GoogleAds\Util\V6;
 
 use Google\Ads\GoogleAds\V6\Enums\AssetFieldTypeEnum\AssetFieldType;
 use Google\Ads\GoogleAds\V6\Enums\ExtensionTypeEnum\ExtensionType;
+use Google\Ads\GoogleAds\V6\Enums\PlaceholderTypeEnum\PlaceholderType;
 use Google\Ads\GoogleAds\V6\Services\AccountBudgetProposalServiceClient;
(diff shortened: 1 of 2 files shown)


fiboknacky (commented):
Could you highlight what new tests you added?

PierrickVoulet:
> Could you highlight what new tests you added?

Sure. Added `testGetNameForTopicView` and `testGetNameForFeedPlaceholderView`.

tests/Google/Ads/GoogleAds/Util/V6/ResourceNamesTest.php
```diff
@@ -2403,6 +2427,31 @@ public function testGetNameForTopicConstant()
...
+        $this->assertEquals($expectedResourceName, ResourceNames::forTopicView(
+            self::CUSTOMER_ID,
+            $adGroupId,
+            $criterionId
+        ));
+
+        $names = TopicViewServiceClient::parseName($expectedResourceName);
+        $this->assertEquals(self::CUSTOMER_ID, $names['customer_id']);
+        $this->assertEquals($adGroupId, $names['ad_group_id']);
+        $this->assertEquals($criterionId, $names['criterion_id']);
```
fiboknacky, line 2452: So, the TopicViewServiceClient generates ad_group_id and criterion_id for us?

PierrickVoulet: I believe so, would not the assert fail if that is not the case? What makes you think that there is something wrong?

fiboknacky: The new scheme sometimes parse a name and generates tokens that are close to the resource name, e.g., 'customer_id' and 'topic_view'.
This case doesn't follow that scheme though.

fiboknacky (approved):

PR #572, opened 2021-04-05
Upgrade GetAdGroupBidModifiers code example

PierrickVoulet:
Editions are made after the [golden](https://github.com/googleads/google-ads-dotnet/blob/master/examples/AdvancedOperations/GetAdGroupBidModifiers.cs).

Change-Id: I9900b47919482907b8a5e8794468eddddf66f8b4

### DIFF
--- a/examples/AdvancedOperations/GetAdGroupBidModifiers.php
+++ b/examples/AdvancedOperations/GetAdGroupBidModifiers.php
@@ -27,7 +27,11 @@ use Google\Ads\GoogleAds\Lib\V6\GoogleAdsClient;
 use Google\Ads\GoogleAds\Lib\V6\GoogleAdsClientBuilder;
 use Google\Ads\GoogleAds\Lib\V6\GoogleAdsException;
 use Google\Ads\GoogleAds\Lib\OAuth2TokenBuilder;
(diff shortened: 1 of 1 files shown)


examples/AdvancedOperations/GetAdGroupBidModifiers.php
```diff
@@ -126,18 +135,53 @@ public static function runExample(
         // the ad group bid modifier in each row.
         foreach ($response->iterateAllElements() as $googleAdsRow) {
             /** @var GoogleAdsRow $googleAdsRow */
+            $AdGroupBidModifier = $googleAdsRow->getAdGroupBidModifier();
```
fiboknacky, line 138: Use the lower-case variable. (I guess this will fail phpcs?)

fiboknacky (changes requested):

examples/AdvancedOperations/GetAdGroupBidModifiers.php
PierrickVoulet: Good catch! It does not make phpcs fail.

fiboknacky (approved):

Issue #523, opened 2021-01-29
Unable to run Laravel Application on Linux/nginx Server 


SubaKrishnan:
Hi,
I had implemented the LaravelSampleApp by adding the extensions, It works well in Local system. But in Live, I have Linux/Nginx server. Here also I had done the same process by installing the extensions & enabled at php.ini file. But gRPC is not working in Nginx Server.

In live, shouts error as 
"gRPC support has been requested but required dependencies have not been found. For details on how to install the gRPC extension please see https://cloud.google.com/php/grpc. "

But, I have gRPC & protobuf at my server 

Also, I had edit the Nginx config file by adding the line "grpc_pass grpc://localhost:50051"

I made a try by referring this document [Nginx Setup](https://www.nginx.com/blog/nginx-1-13-10-grpc/)  But nothing helps. 
Can you guide me on how to do the setup at Linux/Nginx System? 

I would like to know how to configure grpc at Nginx server 

If gRPC is not possible. I would like to know how to implement the LaravelSampleApp using the REST Transport

When I set the transport as rest in google_ads_php.ini. It shows 
"Streaming calls are not supported while using the REST transport"

![image](https://user-images.githubusercontent.com/54890552/106276554-c2866400-625d-11eb-89a8-4b49d3189faf.png)

PierrickVoulet:
Hi @SubaKrishnan,

When the gRPC PHP extension is not available on the system, PHP falls back and uses the gRPC package and REST transport (see full detail [here](https://developers.google.com/google-ads/api/docs/client-libs/php/transport)). The issue is that the REST transport does _not_ support streaming (because it is not HTTP/2 compliant) but the LaravelSampleApp code example _does_ use it by calling the GoogleAdsService.SearchStream method to execute search queries.

One alternative is to change from GoogleAdsService.SearchStream to GoogleAdsService.Search so that your application does not rely on streaming at all but this may lead to other issues, typically higher latency (see full detail [here](https://developers.google.com/google-ads/api/docs/reporting/streaming?hl=en)).

The best would be to fix your Nginx setup so that the gRPC PHP extension is available to the application. I suggest you to contact the Nginx community as they should be much better suited to help you with this. Here are links I could find: [community home](https://www.nginx.com/resources/wiki/community), [get involved](https://www.nginx.com/resources/wiki/community/get_involved).

SubaKrishnan:
Thanks for you help.

I had followed the instructions you gave me. It works well

Issue #311, opened 2020-04-06
$googleAdsServiceClient->search() function is not working


amitklahiri:
Hi. 

**Which version of the client library are you using?**
Latest version, after downloading using composer. 

**Which version of PHP are you using?**
I am using Laravel 7. Also using the artisan server to run the application. 

**Which operating system (Linux, Windows, ...) and version?**
Windows. 

**Actions taken**
 ```
// Generate a refreshable OAuth2 credential for authentication. 
$oAuth2Credential = (new OAuth2TokenBuilder())
->withClientId('XXX123ZZZ123.apps.googleusercontent.com')
->withClientSecret('XXX.....')
->withRefreshToken('XXX................')
->build();

// Construct a Google Ads client configured from a properties file and the OAuth2 credentials above. 
$googleAdsClient = (new GoogleAdsClientBuilder())
->withOAuth2Credential($oAuth2Credential)
->withDeveloperToken('XXX...')
->withLoginCustomerId('XXX...')
->build();
```

**Expected result**
List of campaigns id and name. 

**Actual result**
I am using the below code snippets to display the campaign id and name list. 

```
public function runExample(GoogleAdsClient $googleAdsClient, int $customerId)
    {
        $googleAdsServiceClient = $googleAdsClient->getGoogleAdsServiceClient();

        // Creates a query that retrieves all campaigns.
        $query = 'SELECT campaign.id, campaign.name FROM campaign ORDER BY campaign.id';
        
        // Issues a search stream request.
        /** @var GoogleAdsServerStreamDecorator $stream */
        $stream = $googleAdsServiceClient->search($customerId, $query);

        // Iterates over all rows in all messages and prints the requested field values for
        // the campaign in each row.
        foreach ($stream->readAll() as $response) {
            /** @var SearchGoogleAdsStreamResponse $response */
            foreach ($response->getResults() as $googleAdsRow) {
                /** @var GoogleAdsRow $googleAdsRow */
                printf(
                    "Campaign with ID %d and name '%s' was found.%s",
                    $googleAdsRow->getCampaign()->getId()->getValue(),
                    $googleAdsRow->getCampaign()->getName()->getValue(),
                    PHP_EOL
                );
            }
        }
    }
```

Getting the error in this line of code.
$stream = $googleAdsServiceClient->search($customerId, $query);

Error is as below.
Streaming calls are not supported while using the REST transport

**Anything else we should know about your project / environment**
[...]

Issue #304, opened 2020-03-30
searchStream() is not working


amitklahiri:
Hi. 

I am using the latest version of the client library as available using composer install option. I am getting the below error . 

I am using it with PHP Laravel framework. 

Here is my code, 
   ```
 public static function runExample(GoogleAdsClient $googleAdsClient, int $customerId)
    {
        $googleAdsServiceClient = $googleAdsClient->getGoogleAdsServiceClient();
        // Creates a query that retrieves all campaigns.
        $query = 'SELECT campaign.id, campaign.name FROM campaign ORDER BY campaign.id';
        // Issues a search stream request.
        /** @var GoogleAdsServerStreamDecorator $stream */
        $stream = $googleAdsServiceClient->searchStream($customerId, $query);

        // Iterates over all rows in all messages and prints the requested field values for
        // the campaign in each row.
        foreach ($stream->readAll() as $response) {
            /** @var SearchGoogleAdsStreamResponse $response */
            foreach ($response->getResults() as $googleAdsRow) {
                /** @var GoogleAdsRow $googleAdsRow */
                printf(
                    "Campaign with ID %d and name '%s' was found.%s",
                    $googleAdsRow->getCampaign()->getId()->getValue(),
                    $googleAdsRow->getCampaign()->getName()->getValue(),
                    PHP_EOL
                );
            }
        }
    }
```

It looks like getting the error, in this line,
$stream = $googleAdsServiceClient->searchStream($customerId, $query);

And it occurs because the function searchStream(.....), which seems unavailable. 

What am I supposed to do now? Looking forward to solution ASAP. 

Thanks.

PierrickVoulet:
Hi @amitklahiri - could you provide more information for us to investigate, e.g.:

- OS
- PHP, Protobuf, gRPC versions used
- Protobuf implementation used ([C or PHP](https://developers.google.com/google-ads/api/docs/client-libs/php/protobuf#determine_which_implementation_is_being_used))
- Logs

There are two different methods to search and [streaming](https://developers.google.com/google-ads/api/docs/reporting/streaming) is the one used in this particular code example. Could you give a try with [paging](https://developers.google.com/google-ads/api/docs/reporting/paging) instead? You can refer to the code example [GetAdGroups.php](https://github.com/googleads/google-ads-php/blob/master/examples/BasicOperations/GetAdGroups.php).

fiboknacky:
[...]

PR #576, opened 2021-04-20
Upgrade LaravelSampleApp code example

### LINKED ISSUE #516: Upgrade Laravel in the code example LaravelSampleApp to >=6.20.14
**Describe the problem you are trying to solve**
The [LaravelSampleApp](https://github.com/googleads/google-ads-php/tree/master/examples/LaravelSampleApp) code example depends on Laravel 5.x today which is subject to security issues as per GitHub advisories: [#1](https://github.com/advisories/GHSA-3p32-j457-pg5x), [#2](https://github.com/laravel/framework/security/advisories/GHSA-3p32-j457-pg5x).

**Describe the solution you would like**
To fix the security issue, the minimal version of Laravel should be 6.20.14. The best would be to upgrade straight to 8.x though.

**Additional context**
The security issue has a low severity because it is only used by one of our code examples which does not aim to be production-grade anyway. But it is still an issue because we do not want our users to expand this code example as-is knowing that there is a known security issue.

### LINKED ISSUE #574: Process to implement ads library in laravel
Hello Team,
How I can Implement this google ads library into Laravel. Please send us a guide so that I can follow same and get done my work.

Thanks in Advance 
Rahul

PierrickVoulet:
* Replaced all `GoogleAdsService.SearchStream` calls to `GoogleAdsService.Search` calls. The gRPC PHP Extension is not required anymore to run the example.
* Migrated the version of Laravel from `5.8` to `8.0` (fixes #516).
* Upgraded the version of the library to `8.1` (fixes #574).
* Added the `composer.lock` file so that GitHub's Dependabot can automatically pull-request security updates in the future.
* Excluded the file `google_ads_php.ini` file to avoid unexpected exposure of private data from users.
* Added config rules to make it easier to run unit and code style tests

Change-Id: I3af07376d966328bc2f4bf2609c13a7f3dbc18ef


fiboknacky (commented):
Could you tell why replaced all GoogleAdsService.SearchStream calls with GoogleAdsService.Search calls? This is because it doesn't require the gRPC extension?

PierrickVoulet:
> Could you tell why replaced all GoogleAdsService.SearchStream calls with GoogleAdsService.Search calls? This is because it doesn't require the gRPC extension?

Correct, the purpose is to have it run just fine even when the `grpc` PHP Extension is not installed. This should avoid confusion (see #523, #311, #304).

fiboknacky (approved):

PR #591, opened 2021-05-12
Provide guidance about the loginCustomerId format

akshayagarwal:
The loginCustomerId, if applicable, needs to be specified without dashes, just like the linkedCustomerId. However, while the comment before linkedCustomerId specifies this requirement, the same was not specified before the loginCustomerId. This usually results in a INVALID_LOGIN_CUSTOMER_ID error, without specifying how to remediate the same. Hence, adding guidance about the proper format for the loginCustomerId would be very helpful.


PierrickVoulet (approved):

fiboknacky (approved):

PierrickVoulet:
Thank you for contributing @akshayagarwal !

PR #635, opened 2021-07-29
allow monolog >=1.26

n-gibs:
My org uses Symfony 4.4, which is a long term service release. Currently, google-ads-php >=v7 only allows monolog >=2.0, which is a conflict with Symfony 4.4. With  google-ads-php v6 sunsetting in about a month, we need v8 to allow monolog >=1.26 which is what Symfony 4.4 is locked at. If you want to support Symfony 4.4, this change will need to be made. Thank you.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
     "google/gax": "^1.7.0",
     "grpc/grpc": "^1.36.0",
     "google/protobuf": "^3.17.1",
-    "monolog/monolog": "^2.0 || >=1.26.0"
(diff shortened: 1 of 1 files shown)


n-gibs:
@googlebot I signed it!

PierrickVoulet:
Thank you for raising the concern @n-gibs and for the pull request.

Additional context: Symfony 4.4 will be deprecated in years (see [schedule](https://symfony.com/releases)).

PierrickVoulet (approved):

fiboknacky (approved):

PR #643, opened 2021-09-02
Add library name and version in header

PierrickVoulet:


src/Google/Ads/GoogleAds/Constants.php
```diff
@@ -0,0 +1,31 @@
...
+
+namespace Google\Ads\GoogleAds;
+
+/**
+ * Class to store constants and default values of the library.
+ */
+class Constants
+{
+    // The library name to use in Google's request headers.
+    public const NAME = 'gccl';
```
fiboknacky, line 27: Can we use LIBRARY_NAME here? It's not clear from when it's used (`Constants::NAME`) that this is a library name.

PierrickVoulet: Good point. I did change `VERSION` as well.

fiboknacky (approved):

PR #649, opened 2021-09-15
Fixes #613

### LINKED ISSUE #613: Revert #533 when protocolbuffers/protobuf#8720 is released
#533 was merged to mitigate a log spam issue introduced in newer versions of Protobuf, see #540. It should be reverted as so that warnings are logged again when running unit tests.

PierrickVoulet:
Reverts #533.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,7 @@
     "php": ">=7.3",
     "google/gax": "^1.7.0",
     "grpc/grpc": "^1.36.0",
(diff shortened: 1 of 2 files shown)


fiboknacky (changes requested):
To really fix this issue, I think we should also update the version of protobuf in `composer.json` to point to v3.18.0 too.

fiboknacky (approved):

PR #652, opened 2021-09-17
Fixes #613 (#649)

### LINKED ISSUE #613: Revert #533 when protocolbuffers/protobuf#8720 is released

PierrickVoulet:
* Fixes #613

Reverts #533.

* Upgrade protobuf version constraint to avoid log spamming issue

Change-Id: If747c7cfb612f44c9abd9212a1d5e3284d3cb223

Co-authored-by: pierrick <pierrick@google.com>

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,7 @@
     "php": ">=7.3",
     "google/gax": "^1.7.0",
     "grpc/grpc": "^1.36.0",
-    "google/protobuf": "^3.17.1",
+    "google/protobuf": "^3.18.0",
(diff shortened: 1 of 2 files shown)


PierrickVoulet:
Note: The Kokoro build failure is not due to this change but the edition from `master` to `main`. The job that fails is a QA job and I am working on a fix, please disregard.

fiboknacky (approved):

PR #669, opened 2021-11-01
Catch conversion adjustment failures

ujwaldhakal:
The main changes are

1) Added `conversion_adjustments` key while using `GoogleAdsError` class to get the list of errors

### DIFF
--- a/src/Google/Ads/GoogleAds/Util/V8/GoogleAdsErrors.php
+++ b/src/Google/Ads/GoogleAds/Util/V8/GoogleAdsErrors.php
@@ -25,6 +25,13 @@ use Google\Ads\GoogleAds\V8\Errors\ErrorLocation\FieldPathElement;
 
 final class GoogleAdsErrors
 {
+    private const SUPPORTED_FIELDS = [
+        'operations',
+        'mutate_operations',
+        'conversion_adjustments',
+        'conversions'
+    ];
+
     /**
      * Gets a list of all partial failure error messages for a given response. Operations are
      * indexed from 0.
@@ -109,10 +116,7 @@ final class GoogleAdsErrors
                 $element = $pathElements[0];
                 $fieldName = $element->getFieldName();
                 $index = $element->getIndex();
-                if (
-                    ($fieldName === "operations" || $fieldName === "mutate_operations"
-                    || $fieldName === "conversions" || $fieldName === "conversion_adjustments") && $index == $operationIndex
-                ) {
+                if ( in_array($fieldName, self::SUPPORTED_FIELDS) && $index == $operationIndex) {
                     $result[] = $error;
                 }
             }
--- a/src/Google/Ads/GoogleAds/Util/V8/GoogleAdsErrors.php
+++ b/src/Google/Ads/GoogleAds/Util/V8/GoogleAdsErrors.php
@@ -111,7 +111,7 @@ final class GoogleAdsErrors
                 $index = $element->getIndex();
                 if (
                     ($fieldName === "operations" || $fieldName === "mutate_operations"
-                    || $fieldName === "conversions") && $index == $operationIndex
+                    || $fieldName === "conversions" || $fieldName === "conversion_adjustments") && $index == $operationIndex
                 ) {
                     $result[] = $error;
                 }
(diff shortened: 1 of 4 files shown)


PierrickVoulet:
@ujwaldhakal Thanks for contributing!

Could you provide a piece of code that demonstrates the issue that you are fixing with this PR? I would like to reproduce the issue on my end.

PierrickVoulet (commented):
@ujwaldhakal Thanks for contributing!

Could you provide a piece of code that demonstrates the issue that you are fixing with this PR? I would like to reproduce the issue on my end.

ujwaldhakal:
```
use Google\Ads\GoogleAds\Lib\V8\GoogleAdsClient;
use Google\Ads\GoogleAds\V8\Services\ConversionAdjustmentUploadServiceClient;
use Google\Ads\GoogleAds\V8\Services\ConversionAdjustment;


// this is client after autheticating with oauthBuilder
public function getGoogleAdsClient(): GoogleAdsClient
    {
        return $this->googleAdsClient;
    }
    
    
public function uploadConversionAdjustment() {

/** ConversionAdjustmentUploadServiceClient **/
$client = $this->getGoogleAdsClient()->getConversionAdjustmentUploadServiceClient();

$response = $client->uploadConversionAdjustments(
            $customerId,
     new ConversionAdjustment([//fill data here]), // cli
            true,
        );

```
 now when there is an error and I try to handle partial failures like 
[Handle partial failures](https://github.com/googleads/google-ads-php/blob/d65942e7bcefa91154e234fa91a3b3e5e3fb559e/examples/ErrorHandling/HandlePartialFailure.php#L211
)  

The line [Line](https://github.com/googleads/google-ads-php/blob/d65942e7bcefa91154e234fa91a3b3e5e3fb559e/examples/ErrorHandling/HandlePartialFailure.php#L218) does not print any error because the field conversion adjustment is missing in line [Line](https://github.com/googleads/google-ads-php/blob/d65942e7bcefa91154e234fa91a3b3e5e3fb559e/src/Google/Ads/GoogleAds/Util/V8/GoogleAdsErrors.php#L114)

PierrickVoulet (approved):

PierrickVoulet:
@ujwaldhakal Thanks!

@fiboknacky I was able to reproduce the issue and confirm the fix.

src/Google/Ads/GoogleAds/Util/V8/GoogleAdsErrors.php
```diff
@@ -111,7 +111,7 @@ public static function fromFailure(
                 $index = $element->getIndex();
                 if (
                     ($fieldName === "operations" || $fieldName === "mutate_operations"
-                    || $fieldName === "conversions") && $index == $operationIndex
+                    || $fieldName === "conversions" || $fieldName === "conversion_adjustments") && $index == $operationIndex
```
ujwaldhakal, line 114: @PierrickVoulet is it ok if we cast fieldName as array of fieldName and we can check in array with clean code ?

PierrickVoulet: Not sure I understand correctly your suggestion, could you provide a concrete example or [a GitHub Gist](https://gist.github.com/)?

fiboknacky (commented):
LGTM except for the point that is pending.

src/Google/Ads/GoogleAds/Util/V8/GoogleAdsErrors.php
ujwaldhakal: I mean we can do something like

```
private $supportedFieldNames = ['conversions','conversion_adjustments','operations'];

In line 114 

if(in_array($fieldName, $this->supportedFieldNames) && $index === $operationIndex) {

}

```

PierrickVoulet: Makes sense, the best would be to have it as a constant though.

ujwaldhakal: @PierrickVoulet thanks, i have updated it and when will this be merged and release , since i am implementing custom error handling solutions :)

fiboknacky (approved):

PR #671, opened 2021-11-03
Add gRPC transport credential and insecure channel settings in configuration

PierrickVoulet:


examples/Authentication/google_ads_php.ini
```diff
@@ -46,3 +46,7 @@ refreshToken = "INSERT_OAUTH2_REFRESH_TOKEN_HERE"
 ; Optional transport settings.
 ; By default, "grpc" is used if available otherwise "rest".
 ; transport = "grpc"
+
+; Optional gRPC transport settings.
+; By default, false is used.
```
fiboknacky, line 51: Please add more details on this. For a person with no background about insecure channel, it's hard to understand when and how we can use this option and what the consequence it will have.

src/Google/Ads/GoogleAds/Lib/V7/GoogleAdsClientBuilder.php
```diff
@@ -290,11 +324,41 @@ public function validate()
...
         if (is_null($this->logLevel)) {
             $this->logLevel = LogLevel::INFO;
         } elseif (!defined('Psr\Log\LogLevel::' . strtoupper($this->logLevel))) {
             throw new InvalidArgumentException("The log level must be a valid PSR log level");
         }
+
+        if ($this->grpcChannelIsInsecure && $this->grpcChannelCredential !== null) {
+            throw new InvalidArgumentException(
+                'The gRPC channel credential can only be set when the gRPC channel is set as ' .
+                'not insecure.'
```
fiboknacky, line 342: Can we rename the variable to the "grpcChannelIsSecure" and default it to "true"?
It's error-prone and hard to maintain int the future when using the double negation like this.

src/Google/Ads/GoogleAds/Lib/V7/GoogleAdsClientBuilder.php
```diff
@@ -35,7 +37,10 @@
  */
 final class GoogleAdsClientBuilder extends AbstractGoogleAdsBuilder
 {
+    use GrpcSupportTrait;
+
     private static $DEFAULT_LOGGER_CHANNEL = 'google-ads';
+    private static $DEFAULT_GRPC_CHANNEL_IS_SECURE = true;
```
fiboknacky, line 43: Can we use a new syntax for private const here?

src/Google/Ads/GoogleAds/Lib/V7/GoogleAdsClientBuilder.php
```diff
@@ -80,6 +88,20 @@ public function from(Configuration $configuration)
         );
         $this->proxy = $configuration->getConfiguration('proxy', 'CONNECTION');
         $this->transport = $configuration->getConfiguration('transport', 'CONNECTION');
+        $this->grpcChannelIsSecure =
+            is_null($configuration->getConfiguration('grpcChannelIsSecure', 'CONNECTION'))
+            || $configuration->getConfiguration('grpcChannelIsSecure', 'CONNECTION') === ""
```
fiboknacky, line 93: Can you double check if `getConfiguration` can return null as well? If yes, we probably need to use `empty` instead.

PierrickVoulet: I cannot use `empty` here because `empty(false)` returns `true`. We only want to use the default value when the value is undefined or an empty string.

fiboknacky: Sorry. Somehow, I didn't see line 92 that you checked is_null already.

tests/Google/Ads/GoogleAds/Lib/V7/GoogleAdsClientTest.php
```diff
@@ -132,4 +141,19 @@ public function testNullLinkedCustomerIdNotAppearInClientOptions()
             $googleAdsClient->getGoogleAdsClientOptions()
         );
     }
+
+    public function testGrpcChannelCredentialWhenGrpcChannelIsSecureInClientOptions()
```
fiboknacky, line 145: When GRPC channel is "NOT" secure?
I see you set the value here as false.

fiboknacky (commented):
One minor comment.

tests/Google/Ads/GoogleAds/Lib/V7/GoogleAdsClientTest.php
PierrickVoulet: Good catch!

fiboknacky (approved):

PR #672, opened 2021-11-04
Add gRPC interceptors setting in configuration

### LINKED ISSUE #207: Allow developers write their own interceptors
The current PHP package has a hidden feature - interceptors for grpc requests. 
As seen here: `\Google\Ads\GoogleAds\Lib\V1\ServiceClientFactoryTrait::getGoogleAdsClientOptions`
One application for them is logging (which is already implemented).

For a regular developer, it is not easy to use them and there is no documentation for it available. I feel there is a HUGE benefit in making them available to the general public. For example, I used them to create a VCR layer which would record and playback of all requests and responses (for purposes of testing).

The problem is that `\Google\Ads\GoogleAds\Lib\V1\GoogleAdsClient` is final and doesn't let you customize interceptors.

What do you think?

PierrickVoulet:
Fixes #207


fiboknacky:
This will be probably reviewed after I release support for v9. Could you please wait until that and patch to support v9 too? Thanks.

PierrickVoulet:
> This will be probably reviewed after I release support for v9. Could you please wait until that and patch to support v9 too? Thanks.

Done. Feel free to take a loot whenever you can.

src/Google/Ads/GoogleAds/Lib/ConfigurationTrait.php
```diff
@@ -35,6 +36,7 @@ trait ConfigurationTrait
     private $logLevel;
     private $proxy;
     private $transport;
+    private $grpcInterceptors;
```
fiboknacky, line 39: Same question with the other PR. How do you plan to support this in the config file?

PierrickVoulet: There is no plans to support this in the config file.

We could do it in the future if we decide to support config file in `PHP` format to complement or replace the current `INI` format but this is out of scope of this PR.

fiboknacky: Ditto. Please see the comment in the other PR

fiboknacky (approved):

lezhnev74:
Awesome!

PR #674, opened 2021-11-05
Add middleware settings in configuration

### LINKED ISSUE #207: Allow developers write their own interceptors

PierrickVoulet:
Fixes #207


PierrickVoulet:
@fiboknacky I only implemented the change for V8 as of now. I will replicate for V7 when it looks good to you.

src/Google/Ads/GoogleAds/Lib/ConfigurationTrait.php
```diff
@@ -35,6 +35,8 @@ trait ConfigurationTrait
     private $logLevel;
     private $proxy;
     private $transport;
+    private $unaryMiddlewares;
```
fiboknacky, line 38: I believe we use this class to hold things related to configuration (e.g., those we set in the google_ads_php.ini).
How do you plan to support providing them in the config file?

PierrickVoulet: There is no plans to support this in the config file.

We could do it in the future if we decide to support config file in PHP format to complement or replace the current INI format but this is out of scope of this PR.

fiboknacky: Can you somehow state this anywhere like in the config file itself or as a comment in here?
We've had the 1-1 mapping between the configuration and the config file for years. Deviating from this convention should at least has some explanation.

PierrickVoulet: My plan was to document this in our library doc (DevSite) but I agree that it should be stated in the source code as well, done!

fiboknacky (approved):

Issue #499, opened 2020-12-15
Pagination with 


chrisooo3:
Hi, 
I am trying to do the pagination with google-ads-php like here:
![image](https://user-images.githubusercontent.com/70841270/102247471-4aa7e880-3f00-11eb-87cc-0bc157a72ee5.png)

So this is my code:
```php
    public function runExample(GoogleAdsClient $googleAdsClient, int $customerId)
    {
        $googleAdsServiceClient = $googleAdsClient->getGoogleAdsServiceClient();
        // Creates a query that retrieves all ads.
        $query = "SELECT campaign.name, ad_group.name, "
            . "ad_group_ad.ad.responsive_display_ad.marketing_images, "
            . "ad_group_ad.ad.app_ad.images, ad_group_ad.ad.app_ad.youtube_videos, "
            . "ad_group_ad.ad.responsive_display_ad.youtube_videos, ad_group_ad.ad.local_ad.videos, "
            . "ad_group_ad.ad.video_responsive_ad.videos, ad_group_ad.ad.video_ad.media_file,  "
            . "ad_group_ad.ad.app_engagement_ad.images, ad_group_ad.ad.app_engagement_ad.videos, "
            . "ad_group_ad.ad.display_upload_ad.media_bundle, ad_group_ad.ad.gmail_ad.product_images, "
            . "ad_group_ad.ad.gmail_ad.product_videos, ad_group_ad.ad.gmail_ad.teaser.logo_image, "
            . "ad_group_ad.ad.image_ad.image_url, ad_group_ad.ad.legacy_responsive_display_ad.square_marketing_image, "
            . "ad_group_ad.ad.local_ad.marketing_images, ad_group_ad.ad.responsive_display_ad.logo_images, "
            . "ad_group_ad.ad.responsive_display_ad.square_logo_images, "
            . "ad_group_ad.ad.responsive_display_ad.square_marketing_images, "
            . "ad_group_ad.ad.responsive_display_ad.youtube_videos, "
            . "metrics.impressions, campaign.campaign_budget, campaign.status, "
            . "campaign.start_date, campaign.end_date, metrics.all_conversions, "
            . "metrics.average_cost, ad_group_ad.ad.type, ad_group_ad.ad.id, "
            . "campaign.campaign_budget, metrics.cost_micros, ad_group_ad.status, metrics.impressions "
            . "FROM ad_group_ad "
            . "WHERE segments.date >= '{$this->from}' AND segments.date <= '{$this->to}' "
[...]

PR #696, opened 2021-12-06
Upgrade laravel code example with lazy page loading

PierrickVoulet:
Mitigates #499

### DIFF
--- a/examples/LaravelSampleApp/app/Http/Controllers/GoogleAdsApiController.php
+++ b/examples/LaravelSampleApp/app/Http/Controllers/GoogleAdsApiController.php
@@ -18,13 +18,13 @@
 
 namespace App\Http\Controllers;
 
-use Google\Ads\GoogleAds\Lib\V9\GoogleAdsClient;
+use Google\Ads\GoogleAds\Lib\V8\GoogleAdsClient;
 use Google\Ads\GoogleAds\Util\FieldMasks;
-use Google\Ads\GoogleAds\Util\V9\ResourceNames;
-use Google\Ads\GoogleAds\V9\Enums\CampaignStatusEnum\CampaignStatus;
-use Google\Ads\GoogleAds\V9\Resources\Campaign;
-use Google\Ads\GoogleAds\V9\Services\CampaignOperation;
-use Google\Ads\GoogleAds\V9\Services\GoogleAdsRow;
+use Google\Ads\GoogleAds\Util\V8\ResourceNames;
+use Google\Ads\GoogleAds\V8\Enums\CampaignStatusEnum\CampaignStatus;
+use Google\Ads\GoogleAds\V8\Resources\Campaign;
+use Google\Ads\GoogleAds\V8\Services\CampaignOperation;
+use Google\Ads\GoogleAds\V8\Services\GoogleAdsRow;
 use Illuminate\Http\Request;
 use Illuminate\Pagination\LengthAwarePaginator;
 use Illuminate\View\View;
@@ -35,6 +35,8 @@ class GoogleAdsApiController extends Controller
         'campaign' => ['campaign.id', 'campaign.name', 'campaign.status'],
         'customer' => ['customer.id']
     ];
+
+    private const RESULTS_LIMIT = 100;
 
     /**
      * Controls a POST or GET request submitted in the context of the "Show Report" form.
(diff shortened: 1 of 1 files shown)


examples/LaravelSampleApp/app/Http/Controllers/GoogleAdsApiController.php
```diff
@@ -76,52 +78,112 @@ public function showReportAction(
...
+            // Fetches the page for the latest known page token.
+            $response = $googleAdsClient->getGoogleAdsServiceClient()->search(
+                $customerId,
+                $query,
+                [
+                    'pageSize' => $entriesPerPage,
+                    // Requests to return the total results count. This is necessary to
+                    // determine how many pages of results exist.
+                    'returnTotalResultsCount' => true,
+                    'pageToken' => end($pageTokens)
```
fiboknacky, line 128: Could you add a brief explanation why "end" is used here?

examples/LaravelSampleApp/app/Http/Controllers/GoogleAdsApiController.php
```diff
@@ -76,52 +78,112 @@ public function showReportAction(
...
+                ]
+            );
+            if ($response->getPage()->getNextPageToken()) {
+                // Adds the new page token found.
+                $pageTokens[] = $response->getPage()->getNextPageToken();
```
fiboknacky, line 133: Wouldn't this do "two" next page tokens?

examples/LaravelSampleApp/app/Http/Controllers/GoogleAdsApiController.php
```diff
@@ -76,52 +78,112 @@ public function showReportAction(
...
+            $pageTokens = $request->session()->get('pageTokens');
         }
 
+        // Determines the number of the page to load (the first one by default).
         $pageNo = $request->input('page') ?: 1;
 
-        // Creates a length aware paginator to supply a given page of results for the view, based on
-        // the collection of results, the page number and the specified number of entries per page.
+        // Fetches pages in sequence to find the page token of the requested page if unknown.
+        while (count($pageTokens) < $pageNo) {
```
fiboknacky, line 118: I have a trouble understanding the purpose of this loop, partly because of the if clause at the end and the `end($pageTokens)` as well.

Can you elaborate more on what to expect at the beginning of this loop too?
The loop seems to go backward? From the end of the `$pageTokens` array? 
Is there any particular reason why we do like this instead of going forward?

fiboknacky: Or is this a kind of pre-processing?

examples/LaravelSampleApp/app/Http/Controllers/GoogleAdsApiController.php
```diff
@@ -76,52 +78,112 @@ public function showReportAction(
...
+                // Requests to return the total results count. This is necessary to
+                // determine how many pages of results exist.
+                'returnTotalResultsCount' => true,
+                'pageToken' => $pageTokens[$pageNo - 1]
+            ]
+        );
+        if ($response->getPage()->getNextPageToken()) {
+            // Adds the new page token if there is one (it is not the last page).
+            $pageTokens[] = $response->getPage()->getNextPageToken();
+        }
```
fiboknacky, line 156: Same question: could you double-check if this will advance the pointer twice?

fiboknacky (changes requested):

PierrickVoulet:
@fiboknacky I improved comments based on your feedback. Let me know if it is clearer now.

examples/LaravelSampleApp/app/Http/Controllers/GoogleAdsApiController.php
PierrickVoulet: The code example works as intended, there is no "pointer" used here per se. What we do is to fetch pages based on which one has to be displayed and store page tokens along the way for potential future use.

examples/LaravelSampleApp/app/Http/Controllers/GoogleAdsApiController.php
```diff
@@ -76,52 +78,120 @@ public function showReportAction(
...
+                // that it can be reused later if necessary.
+                $pageTokens[] = $response->getPage()->getNextPageToken();
+            } else {
+                // Otherwise changes the requested page number for the latest page that we have
+                // fetched until now, the requested page number was invalid.
+                $pageNo = count($pageTokens);
+            }
+        }
+
+        // Fetches the actual page to display results of.
```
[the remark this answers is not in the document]
PierrickVoulet: Should be better now, let me know.

fiboknacky (approved):

PR #698, opened 2021-12-10
Changes for release Dec 2021

fiboknacky:
-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,23 @@
+## 13.0.0
+*   Updated `composer.json` to require the minimum PHP version of 7.4.
+*   Added support for setting gRPC channel, gRPC interceptor and middleware.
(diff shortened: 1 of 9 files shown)


PierrickVoulet:
Please update [the version of the client library](https://github.com/googleads/google-ads-php/blob/main/src/Google/Ads/GoogleAds/Constants.php#L30).

PierrickVoulet (changes requested):
Please update [the version of the client library](https://github.com/googleads/google-ads-php/blob/main/src/Google/Ads/GoogleAds/Constants.php#L30).

fiboknacky:
> Please update [the version of the client library](https://github.com/googleads/google-ads-php/blob/main/src/Google/Ads/GoogleAds/Constants.php#L30).
Done.

PierrickVoulet (approved):

PR #702, opened 2021-12-15
Remove printing of GCLID in UploadConversionWithIdentifiers

jradcliff:
GCLID is not used in this example, so will always be absent.

### DIFF
--- a/examples/Remarketing/UploadConversionWithIdentifiers.php
+++ b/examples/Remarketing/UploadConversionWithIdentifiers.php
@@ -193,10 +193,9 @@ class UploadConversionWithIdentifiers
             // Only prints valid results.
             if ($clickConversionResult->hasGclid()) {
                 printf(
-                    "Uploaded conversion that occurred at '%s' from Google Click ID '%s' to "
+                    "Uploaded conversion that occurred at '%s' to "
                     . "'%s'.%s",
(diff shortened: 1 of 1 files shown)


jradcliff:
Credit to @AnashOommen for catching this in the Java example, which I'll fix as well.

PierrickVoulet:
Kokoro build failure will be fixed with #703.

fiboknacky (approved):

fiboknacky (changes requested):
Could you move line 197 to line 196 too? That should fit in one line.

jradcliff:
Combined lines as requested

fiboknacky (commented):
Sorry for not being clear. I meant:

```php
                printf(
                    "Uploaded conversion that occurred at '%s' to '%s'.%s",
                    $clickConversionResult->getConversionDateTime(),
                    $clickConversionResult->getConversionAction(),
                    PHP_EOL
                );
```

BTW, in this case, would the condition in the line 194 be still okay?
It's still checking `hasGclid()`.

jradcliff:
Good catch. Changed to check the conversion date time, since that's a required field. Doing the same in the updated Java example at https://github.com/googleads/google-ads-java/pull/541/files.

fiboknacky (approved):

PR #718, opened 2022-01-19
Fix #717

### LINKED ISSUE #717: GoogleAdsResponseMetadata getRequestId() is always null for REST transport
**Your client library and Google Ads API versions:**

* Client library version: `v12.1.0`
* Google Ads API version: `V9`

**Your environment:**

* The PHP version: 7.3.15
* The name and version of the operating system: Linux 95b4c8e933d3 5.15.8-76051508-generic #202112141040\~1639505278\~21.04\~0ede46a-Ubuntu SMP Fri Dec 17 19: x86_64
* Whether the PHP Extension `protobuf` is used or not: not installed
* Whether the PHP Extension `grpc` is used or not: not installed
* The type of transport being used (`grpc` or `rest`): REST

**Description of the bug:**
The `GoogleAdsResponseMetadata` class's `getRequestId()` method does return `null` for REST requests. Inspecting the actual headers that are passed to the constructor shows that the headers are upper case (i.e. `Request-Id`) whereas the `GoogleAdsMetadataTrait` expects them to be lower case (i.e. `request-id`).

**Steps to reproduce:**
```php
$accountId = 'REDACTED';
$serviceClient = $this->getClient()->getGoogleAdsServiceClient();

$query = "SELECT conversion_action.id,conversion_action.name FROM conversion_action";

/** @var GoogleAdsResponseMetadata $metadata */
[$response, $metadata] = $serviceClient->search($accountId, $query, ['withResponseMetadata' => true]);

var_dump($metadata->getRequestId()); // returns `null` 
var_dump($metadata);
```
var_dump of $metadata:
```
object(Google\Ads\GoogleAds\Lib\V9\GoogleAdsResponseMetadata)[34231]
  private 'metadata' => 
    array (size=12)
      'Request-Id' => 
        array (size=1)
          0 => string 'sA7BMapJTIRl3goJv0rZUQ' (length=22)
      'Content-Type' => 
        array (size=1)
          0 => string 'application/json; charset=UTF-8' 
....
```

**Expected behavior:**
`getRequestId()` returns the request's id.

**Anything else we should know about your project / environment:**
I am developing against a "Test account" using the REST transport. I am not sure if this is a general issue or specific to the transport, account or my environment.

PierrickVoulet:
Change-Id: Ib9f216260aca336650dcf1b6fe141463d8dc2381

### DIFF
--- a/src/Google/Ads/GoogleAds/Lib/V7/GoogleAdsMetadataTrait.php
+++ b/src/Google/Ads/GoogleAds/Lib/V7/GoogleAdsMetadataTrait.php
@@ -39,8 +39,11 @@ trait GoogleAdsMetadataTrait
      */
     private function getFirstHeaderValue($key, $metadata)
     {
-        if (isset($metadata[$key])) {
-            $valueArray = $metadata[$key];
+        // HTTP header field names are case-insensitive.
+        $lowerMetadata = array_change_key_case($metadata);
+        $lowerKey = strtolower($key);
+        if (isset($lowerMetadata[$lowerKey])) {
+            $valueArray = $lowerMetadata[$lowerKey];
             if (count($valueArray) > 0) {
                 return $valueArray[0];
             }
--- a/src/Google/Ads/GoogleAds/Lib/V8/GoogleAdsMetadataTrait.php
+++ b/src/Google/Ads/GoogleAds/Lib/V8/GoogleAdsMetadataTrait.php
@@ -39,8 +39,11 @@ trait GoogleAdsMetadataTrait
      */
     private function getFirstHeaderValue($key, $metadata)
     {
-        if (isset($metadata[$key])) {
-            $valueArray = $metadata[$key];
+        // HTTP header field names are case-insensitive.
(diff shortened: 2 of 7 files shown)


PierrickVoulet:
@fiboknacky Please review the change in V9, I will replicate the changes to remaining versions when done.

fiboknacky (commented):
Looks like you're lowering the cases of the keys twice, one in `getFirstHeaderValue` and another in `getRequestId`.
Is there any reason for this?

I ask because it seems we have only one additional test here, and doing that in one place would be enough?

PierrickVoulet:
> Looks like you're lowering the cases of the keys twice, one in `getFirstHeaderValue` and another in `getRequestId`. Is there any reason for this?
> 
> I ask because it seems we have only one additional test here, and doing that in one place would be enough?

Thanks for your feedback. I edited the logic to be self-contained in `getFirstHeaderValue` which is much better because as generic as it can be.

fiboknacky (approved):

PR #729, opened 2022-02-01
Add code example NavigateSearchResultPagesCachingTokens

PierrickVoulet:


examples/Misc/NavigateSearchResultPagesCachingTokens.php
```diff
@@ -0,0 +1,270 @@
...
+                $googleAdsRow->getCampaign()->getId(),
+                $googleAdsRow->getCampaign()->getName(),
+                PHP_EOL
+            );
+        }
+    }
+    // [END navigate_search_result_pages_caching_tokens]
+
+    /**
+     * Update the cache of page tokens based on a page that was retrieved.
```
fiboknacky, line 253: "Updates..."

examples/Misc/NavigateSearchResultPagesCachingTokens.php
```diff
@@ -0,0 +1,270 @@
...
+     *
+     * @param array &$pageTokens the cache of page tokens to update
+     * @param Page $page the page that was retrieved
+     * @param int $pageNumber the number of the page that was retrieved
+     */
+    private static function cacheNextPageToken(array &$pageTokens, Page $page, int $pageNumber)
+    {
+        if ($page->getNextPageToken() && !isset($pageTokens[$pageNumber])) {
+            // Update the cache with the next page token if it is not set yet.
```
fiboknacky, line 262: Updates...

examples/Misc/NavigateSearchResultPagesCachingTokens.php
```diff
@@ -0,0 +1,270 @@
...
+            self::RESULTS_LIMIT,
+            $response->getPage()->getResponseObject()->getTotalResultsCount()
+        );
+        printf('Total number of campaigns found: %d.%s', $totalNumberOfResults, PHP_EOL);
+
+        // Determines the total number of pages and prints it.
+        $totalNumberOfPages = ceil($totalNumberOfResults / self::PAGE_SIZE);
+        printf('Total number of pages: %d.%s', $totalNumberOfPages, PHP_EOL);
+        if (!$totalNumberOfPages) {
+            throw new Exception(
```
fiboknacky, line 155: I think this can be just a normal print and exit the method unless we have a specific reason to throw an error instead.

examples/Misc/NavigateSearchResultPagesCachingTokens.php
```diff
@@ -0,0 +1,270 @@
...
+                'Could not find any results.'
+            );
+        }
+
+        $middlePageNumber = ceil($totalNumberOfPages / 2);
```
fiboknacky, line 160: Can we explain why `$middlePageNumber` is used here? I think you probably want to demonstrate how to navigate forward and backward, but it wasn't clear when I read until this line. Readers could forget what it's about too.

examples/Misc/NavigateSearchResultPagesCachingTokens.php
```diff
@@ -0,0 +1,270 @@
...
+        }
+    }
+
+    // [START navigate_search_result_pages_caching_tokens]
+    /**
+     * Fetches and prints the results of a page of a search using a cache of page tokens.
+     *
+     * @param GoogleAdsServiceClient $googleAdsServiceClient the Google Ads API Service client
+     * @param int $customerId the customer ID
+     * @param string $searchQuery  the search query
```
fiboknacky, line 191: extra space after `$searchQuery`

examples/Misc/NavigateSearchResultPagesCachingTokens.php
```diff
@@ -0,0 +1,270 @@
...
+     */
+    private static function fetchAndPrintPageResults(
+        GoogleAdsServiceClient $googleAdsServiceClient,
+        int $customerId,
+        string $searchQuery,
+        array $searchOptions,
+        int $pageNumber,
+        array &$pageTokens
+    ) {
+        // There is no need to fetch the pages we already know the page tokens for.
```
fiboknacky, line 204: I think the comment doesn't quite 100% match what comes after it.
I'd swap the if clause, using `isset` instead. This will align with the Google recommendation that prefers positive clauses over the negative ones.

fiboknacky (changes requested):

examples/Misc/NavigateSearchResultPagesCachingTokens.php
PierrickVoulet: I changed for a normal print because throwing an exception does not bring any value but I use `exit(1)` instead of `return` because I think we should consider the code example to fail when this happens.

fiboknacky (changes requested):

examples/Misc/NavigateSearchResultPagesCachingTokens.php
```diff
@@ -0,0 +1,273 @@
...
+        if (isset($pageTokens[$pageNumber - 1])) {
+            printf(
+                'The token of the requested page was cached, we will use it to get the results.%s',
+                PHP_EOL
+            );
+            $currentPageNumber = $pageNumber;
+            $currentPageNumber = count($pageTokens);
```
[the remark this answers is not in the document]
PierrickVoulet: Good catch!

fiboknacky (approved):

AnashOommen (approved):

PR #756, opened 2022-03-07
Added AddPerformanceMaxProductListingGroupTree example

fiboknacky:
This is based on https://github.com/googleads/google-ads-dotnet/blob/main/examples/ShoppingAds/AddPerformanceMaxProductListingGroupTree.cs.

But I don't use the factory pattern like in the .NET library, as it would clutter the implementation.
The current version should be close to the AddShoppingProductListingGroupTree example, so the users should be familiar with this implementation.

### DIFF
--- a/examples/ShoppingAds/AddPerformanceMaxProductListingGroupTree.php
+++ b/examples/ShoppingAds/AddPerformanceMaxProductListingGroupTree.php
@@ -130,6 +130,7 @@ class AddPerformanceMaxProductListingGroupTree
         }
     }
 
+    // [START add_performance_max_product_listing_group_tree]
     /**
      * Runs the example.
      *
@@ -263,6 +264,7 @@ class AddPerformanceMaxProductListingGroupTree
         self::printResponseDetails($mutateOperations, $response);
         // [END add_performance_max_product_listing_group_tree_1]
     }
+    // [END add_performance_max_product_listing_group_tree]
 
     // [START add_performance_max_product_listing_group_tree_7]
     /**
@@ -280,6 +282,9 @@ class AddPerformanceMaxProductListingGroupTree
     ): array {
         $googleAdsServiceClient = $googleAdsClient->getGoogleAdsServiceClient();
         // Creates a query that retrieves asset group listing group filters.
+        // The limit to the number of listing group filters permitted in a Performance
+        // Max campaign can be found here:
+        // https://developers.google.com/google-ads/api/docs/best-practices/system-limits.
         $query = sprintf(
             'SELECT asset_group_listing_group_filter.resource_name, '
             . 'asset_group_listing_group_filter.parent_listing_group_filter '
@@ -339,8 +344,7 @@ class AddPerformanceMaxProductListingGroupTree
             $parentResourceName = $assetGroupListingGroupFilter->getParentListingGroupFilter();
             $siblings = [];
 
-            // Checks to see if we've already visited a sibling in this group, and fetches or
-            // creates a new set as required.
+            // Checks to see if we've already visited a sibling in this group and fetches it.
             if (array_key_exists($parentResourceName, $parentsToChildren)) {
                 $siblings = $parentsToChildren[$parentResourceName];
             }
@@ -427,9 +431,9 @@ class AddPerformanceMaxProductListingGroupTree
 
         return new MutateOperation([
             'asset_group_listing_group_filter_operation'
-            => new AssetGroupListingGroupFilterOperation([
-                'create' => $assetGroupListingGroupFilter
-            ])
+                => new AssetGroupListingGroupFilterOperation([
+                    'create' => $assetGroupListingGroupFilter
+                ])
         ]);
     }
     // [END add_performance_max_product_listing_group_tree_4]
--- a/examples/ShoppingAds/AddPerformanceMaxProductListingGroupTree.php
+++ b/examples/ShoppingAds/AddPerformanceMaxProductListingGroupTree.php
@@ -34,6 +34,8 @@ use Google\Ads\GoogleAds\V10\Enums\ListingGroupFilterVerticalEnum\ListingGroupFi
 use Google\Ads\GoogleAds\V10\Errors\GoogleAdsError;
 use Google\Ads\GoogleAds\V10\Resources\AssetGroupListingGroupFilter;
 use Google\Ads\GoogleAds\V10\Resources\ListingGroupFilterDimension;
+use Google\Ads\GoogleAds\V10\Resources\ListingGroupFilterDimension\ProductBrand;
+use Google\Ads\GoogleAds\V10\Resources\ListingGroupFilterDimension\ProductCondition;
 use Google\Ads\GoogleAds\V10\Services\AssetGroupListingGroupFilterOperation;
 use Google\Ads\GoogleAds\V10\Services\GoogleAdsRow;
 use Google\Ads\GoogleAds\V10\Services\MutateGoogleAdsResponse;
@@ -187,7 +189,7 @@ class AddPerformanceMaxProductListingGroupTree
             $tempId--,
             self::LISTING_GROUP_ROOT_TEMPORARY_ID,
             new ListingGroupFilterDimension([
-                'product_condition' => new ListingGroupFilterDimension\ProductCondition([
+                'product_condition' => new ProductCondition([
                     'condition' => ListingGroupFilterProductCondition::PBNEW
                 ])
             ])
@@ -199,7 +201,7 @@ class AddPerformanceMaxProductListingGroupTree
             $tempId--,
             self::LISTING_GROUP_ROOT_TEMPORARY_ID,
             new ListingGroupFilterDimension([
-                'product_condition' => new ListingGroupFilterDimension\ProductCondition([
+                'product_condition' => new ProductCondition([
                     'condition' => ListingGroupFilterProductCondition::USED
                 ])
             ])
@@ -218,7 +220,7 @@ class AddPerformanceMaxProductListingGroupTree
             new ListingGroupFilterDimension([
                 // All sibling nodes must have the same dimension type. We use an empty
                 // ProductCondition to indicate that this is an "Other" partition.
-                'product_condition' => new ListingGroupFilterDimension\ProductCondition()
+                'product_condition' => new ProductCondition()
             ])
         );
 
@@ -227,11 +229,9 @@ class AddPerformanceMaxProductListingGroupTree
             $assetGroupId,
             $tempId--,
             $conditionOtherSubdivisionId,
-            new ListingGroupFilterDimension([
-                'product_brand' => new ListingGroupFilterDimension\ProductBrand([
-                    'value' => 'CoolBrand'
-                ])
-            ])
+            new ListingGroupFilterDimension(
+                ['product_brand' => new ProductBrand(['value' => 'CoolBrand'])]
+            )
         );
 
         $mutateOperations[] = self::createMutateOperationForUnit(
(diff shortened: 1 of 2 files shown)


Kenimartonia:
Thanks and best regards

examples/ShoppingAds/AddPerformanceMaxProductListingGroupTree.php
```diff
@@ -0,0 +1,602 @@
...
+        foreach ($response->iterateAllElements() as $googleAdsRow) {
+            /** @var GoogleAdsRow $googleAdsRow */
+            $assetGroupListingGroupFilters[] = $googleAdsRow->getAssetGroupListingGroupFilter();
+        }
+        return $assetGroupListingGroupFilters;
+    }
+    // [END add_performance_max_product_listing_group_tree_7]
+
+    /**
+     * Removes an existing tree of asset group listing group filters.
```
PierrickVoulet, line 312: It does not remove them, it only creates the mutate operations to do so.

Please replace with `Creates mutate operations for removing ...`

examples/ShoppingAds/AddPerformanceMaxProductListingGroupTree.php
```diff
@@ -0,0 +1,602 @@
...
+     *
+     * Asset group listing group filters must be removed in a specific order: all of the children
+     * of a filter must be removed before the filter itself, otherwise the API will return an
+     * error.
+     *
+     * @param GoogleAdsClient $googleAdsClient the Google Ads API client
```
PierrickVoulet, line 318: Both the parameters `$googleAdsClient` and `$customerId` do not exist.

examples/ShoppingAds/AddPerformanceMaxProductListingGroupTree.php
```diff
@@ -0,0 +1,602 @@
...
+     * @param int $customerId the customer ID
+     * @param AssetGroupListingGroupFilter[] $assetGroupListingGroupFilters the existing asset
+     *     group listing group filters
+     * @return MutateOperation[] the list of MutateOperations to remove all listing groups
+     */
+    private static function createMutateOperationsForRemovingListingGroupFiltersTree(
+        array $assetGroupListingGroupFilters
+    ): array {
+        if (empty($assetGroupListingGroupFilters)) {
```
PierrickVoulet, line 327: This seems unnecessary (we only call this method when there are in the main method).

examples/ShoppingAds/AddPerformanceMaxProductListingGroupTree.php
```diff
@@ -0,0 +1,602 @@
...
+            throw new \UnexpectedValueException('No listing group filters to remove');
+        }
+
+        $resourcesToListingGroupFilters = [];
```
PierrickVoulet, line 331: `resourceNamesToListingGroupFilters`

examples/ShoppingAds/AddPerformanceMaxProductListingGroupTree.php
```diff
@@ -0,0 +1,602 @@
...
+            $parentResourceName = $assetGroupListingGroupFilter->getParentListingGroupFilter();
+            $siblings = [];
+
+            // Checks to see if we've already visited a sibling in this group, and fetches or
+            // creates a new set as required.
+            if (array_key_exists($parentResourceName, $parentsToChildren)) {
+                $siblings = $parentsToChildren[$parentResourceName];
+            }
+            $siblings[] = $assetGroupListingGroupFilter->getResourceName();
+            $parentsToChildren[$parentResourceName] = $siblings;
```
PierrickVoulet, line 356: This is only necessary when executed in an `else` clause of the condition above

examples/ShoppingAds/AddPerformanceMaxProductListingGroupTree.php
```diff
@@ -0,0 +1,602 @@
...
+     * factory's asset group.
+     *
+     * The root node or partition is the default, which is displayed as "All Products".
+     *
+     * @param int $customerId the customer ID
+     * @param int $assetGroupId the asset group ID
+     * @param int $rootListingGroupId the root listing group ID
+     * @return MutateOperation the mutate operation for creating the root
+     */
+    // [START add_performance_max_product_listing_group_tree_4]
```
PierrickVoulet, line 415: The golden includes the function comment (not sure if intended but let's keep it consistent)

examples/ShoppingAds/AddPerformanceMaxProductListingGroupTree.php
```diff
@@ -0,0 +1,602 @@
...
+    private static function createMutateOperationForRoot(
```
PierrickVoulet, line 416: This function seems to be unnecessary, should we refactor to remove it altogether?

If we want to keep it, I suggest duplicating the content of `createMutateOperationForSubdivision` and removing the support of the special case for root in `createMutateOperationForSubdivision` for easier understanding.

examples/ShoppingAds/AddPerformanceMaxProductListingGroupTree.php
```diff
@@ -0,0 +1,602 @@
...
+     * @param ListingGroupFilterDimension|null $listingGroupFilterDimension the listing group
+     *     filter dimension to associate with the asset group listing group filter
+     * @return MutateOperation the mutate operation for creating a subdivision
+     */
+    // [START add_performance_max_product_listing_group_tree_5]
+    private static function createMutateOperationForSubdivision(
+        int $customerId,
+        int $assetGroupId,
+        int $assetGroupListingGroupFilterId,
+        ?int $parentId = null,
```
PierrickVoulet, line 447: `= null` can be removed (`?` is enough). Same thing for the next parameter in the list.

examples/ShoppingAds/AddPerformanceMaxProductListingGroupTree.php
```diff
@@ -0,0 +1,602 @@
...
+     * @param int $assetGroupId the asset group ID
+     * @param int $assetGroupListingGroupFilterId the ID of the asset group listing group filter to
+     *     be created
+     * @param int|null $parentId the ID of the parent of asset group listing group filter to be
+     *      created
+     * @param ListingGroupFilterDimension|null $listingGroupFilterDimension the listing group
+     *     filter dimension to associate with the asset group listing group filter
+     * @return MutateOperation the mutate operation for creating a subdivision
+     */
+    // [START add_performance_max_product_listing_group_tree_5]
```
PierrickVoulet, line 442: ditto: The golden includes the function comment (not sure if intended but let's keep it consistent)

examples/ShoppingAds/AddPerformanceMaxProductListingGroupTree.php
```diff
@@ -0,0 +1,602 @@
...
+     * @param int $assetGroupId the asset group ID
+     * @param int $assetGroupListingGroupFilterId the ID of the asset group listing group filter to
+     *     be created
+     * @param int $parentId the ID of the parent of asset group listing group filter to be
+     *      created
+     * @param ListingGroupFilterDimension $listingGroupFilterDimension the listing group
+     *     filter dimension to associate with the asset group listing group filter
+     * @return MutateOperation the mutate operation for creating a unit
+     */
+    // [START add_performance_max_product_listing_group_tree_6]
```
PierrickVoulet, line 504: ditto: The golden includes the function comment (not sure if intended but let's keep it consistent)

PierrickVoulet (changes requested):

examples/ShoppingAds/AddPerformanceMaxProductListingGroupTree.php
```diff
@@ -0,0 +1,602 @@
...
+            'SELECT asset_group_listing_group_filter.resource_name, '
+            . 'asset_group_listing_group_filter.parent_listing_group_filter '
+            . 'FROM asset_group_listing_group_filter '
+            . 'WHERE asset_group_listing_group_filter.asset_group = "%s"',
+            $assetGroupResourceName
+        );
+
+        // Issues a search request by specifying page size.
+        $response =
+            $googleAdsServiceClient->search($customerId, $query, ['pageSize' => self::PAGE_SIZE]);
```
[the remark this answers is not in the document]
fiboknacky: I think it's fine for both ways. The golden uses paging so I follow it for consistency.

examples/ShoppingAds/AddPerformanceMaxProductListingGroupTree.php
```diff
@@ -0,0 +1,602 @@
...
+     * and uses it to extract the new entity's name and resource name.
+     *
+     * @param MutateOperation[] $mutateOperations the submitted mutate operations
+     * @param MutateGoogleAdsResponse $mutateGoogleAdsResponse the mutate Google Ads response
+     */
+    private static function printResponseDetails(
+        array $mutateOperations,
+        MutateGoogleAdsResponse $mutateGoogleAdsResponse
+    ): void {
+        $i = 0;
```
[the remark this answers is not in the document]
fiboknacky: Done. I need to change the variable name a bit, or I cannot make it pass PSR-12 without having line length over 100 chars.

examples/ShoppingAds/AddPerformanceMaxProductListingGroupTree.php
```diff
@@ -0,0 +1,602 @@
...
+    // [END add_performance_max_product_listing_group_tree_6]
+
+    /**
+     * Prints the details of a mutate google ads response. Parses the "response" oneof field name
+     * and uses it to extract the new entity's name and resource name.
+     *
+     * @param MutateOperation[] $mutateOperations the submitted mutate operations
+     * @param MutateGoogleAdsResponse $mutateGoogleAdsResponse the mutate Google Ads response
+     */
+    private static function printResponseDetails(
```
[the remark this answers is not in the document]
fiboknacky: I think you need to weigh that with consistency. The golden CL does this.
[AddPerformanceMaxCampaign](https://github.com/googleads/google-ads-php/blob/main/examples/AdvancedOperations/AddPerformanceMaxCampaign.php#L648) also does something similar to this file, though it's a bit simpler.

examples/ShoppingAds/AddPerformanceMaxProductListingGroupTree.php
fiboknacky: I don't think so. Can you tell why you think like that?

examples/ShoppingAds/AddPerformanceMaxProductListingGroupTree.php
```diff
@@ -412,40 +405,54 @@ private static function createMutateOperationsForRemovingDescendents(
...
+            // another sub-tree.
+            'type' => ListingGroupFilterType::SUBDIVISION,
+            // Because this is a Performance Max campaign for retail, we need to specify
+            // that this is in the shopping vertical.
+            'vertical' => ListingGroupFilterVertical::SHOPPING
+        ]);
+
+        return new MutateOperation([
+            'asset_group_listing_group_filter_operation'
+            => new AssetGroupListingGroupFilterOperation([
```
PierrickVoulet, line 430: Indentation is not perfect.

examples/ShoppingAds/AddPerformanceMaxProductListingGroupTree.php
PierrickVoulet: It works but I find the logic a bit confusing.

I would expect that we only have to execute `$parentsToChildren[$parentResourceName] = $siblings` with `$siblings` a new array when `$parentsToChildren[$parentResourceName]` is unset.

I guess is it nit-picking so we can keep it as-is.

PierrickVoulet (changes requested):

PierrickVoulet (approved):

wfansh (approved):

PR #769, opened 2022-03-18
Updated PMax campaign examples 

fiboknacky:
To add asset group signals and asset group listing group.

### DIFF
--- a/src/Google/Ads/GoogleAds/Util/V10/ResourceNames.php
+++ b/src/Google/Ads/GoogleAds/Util/V10/ResourceNames.php
@@ -105,6 +105,25 @@ use Google\ApiCore\PathTemplate;
  */
 final class ResourceNames
 {
+
+    /**
+     * Generates a resource name of audience.
+     *
+     * @param string $customerId
+     * @param string $audienceId
+     * @return string the audience resource name
+     */
+    public static function forAudience(
+        $customerId,
+        $audienceId
+    ): string {
+        $pathTemplate = new PathTemplate('customers/{customer_id}/audiences/{audience_id}');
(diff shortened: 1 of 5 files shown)


src/Google/Ads/GoogleAds/Util/V10/ResourceNames.php
```diff
@@ -105,6 +105,25 @@
...
 {
+
+    /**
+     * Generates a resource name of audience.
+     *
+     * @param string $customerId
+     * @param string $audienceId
+     * @return string the audience resource name
+     */
+    public static function forAudience(
```
PierrickVoulet, line 116: @fiboknacky Did you update our code generator so that this will be generated in the next major version?

tests/Google/Ads/GoogleAds/Util/V10/ResourceNamesTest.php
```diff
@@ -108,6 +108,20 @@
  */
 class ResourceNamesTest extends TestCase
 {
+    /**
+     * @covers \Google\Ads\GoogleAds\Util\V10\ResourceNames::forAudience()
+     */
+    public function testGetNameForAudience()
```
PierrickVoulet, line 114: @fiboknacky Did you add this manually? If so, did you update our code generator so that this will be generated in the next major version?

PierrickVoulet (approved):

src/Google/Ads/GoogleAds/Util/V10/ResourceNames.php
fiboknacky: I'll do this soon.

PR #808, opened 2022-06-02
Add support for Monolog 3

willemverspyck:


composer.json
```diff
@@ -7,7 +7,7 @@
     "google/gax": "^1.7.0",
     "grpc/grpc": "^1.36.0",
     "google/protobuf": "^3.19.4",
-    "monolog/monolog": "^1.26 || ^2.0"
+    "monolog/monolog": "^1.26 || ^2.0 || ^3.0"
   },
```
fiboknacky, line 11: Could you please also remove ^1.26 ? We don't need that anymore. 
Monolog 2.0 requires PHP 7.2, which is lower than the min PHP version we require right now, so supporting Monolog 2.0 should be enough.

fiboknacky (changes requested):

fiboknacky (approved):

PR #814, opened 2022-06-15
Changes for release v11.

fiboknacky:


PierrickVoulet (approved):

PierrickVoulet (commented):
Please wait before releasing.

PierrickVoulet (changes requested):
Please wait before releasing

fiboknacky:
Note that I added two more commits to this PR. Please review them separately.

src/Google/Ads/GoogleAds/Lib/V10/GoogleAdsCallLogger.php
```diff
@@ -36,8 +35,16 @@ final class GoogleAdsCallLogger
     private $endpoint;
     private $logMessageFormatter;
     private $context;
-    private static $logLevelList;
-    private static $logLevelNamesToNormalizedValues;
+    private const LOG_LEVELS  = [
```
PierrickVoulet, line 38: Could you add a comment to explain what these key/values represent?

My understanding is that the keys are all the log levels registered under [RFC 5424](http://tools.ietf.org/html/rfc5424) and the values are the index of severity in increasing order.

If that is the case, what about using an ordered list instead? I think it would makes things easier to understand and implement.

PierrickVoulet (changes requested):

src/Google/Ads/GoogleAds/Lib/V10/GoogleAdsCallLogger.php
```diff
@@ -35,6 +35,10 @@ final class GoogleAdsCallLogger
     private $endpoint;
     private $logMessageFormatter;
     private $context;
+    // An associative array of all the log levels (based on the PSR-7 standard) to sequence numbers.
```
PierrickVoulet, line 38: I think you mean the PSR-3 standard?

https://www.php-fig.org/psr/psr-3/

PierrickVoulet (changes requested):

src/Google/Ads/GoogleAds/Lib/V10/GoogleAdsCallLogger.php
fiboknacky: Sorry. Done.

PierrickVoulet (approved):

PR #847, opened 2022-09-29
Add LOGIN_CUSTOMER_ID

andresayac:
https://developers.google.com/google-ads/api/docs/concepts/call-structure#cid%22


fiboknacky:
Login customer ID can be applied to all examples, but not in this way. Even if you add this to the main function, it's not going to work.

PR #887, opened 2023-03-29
Added AddThingsToDoAd example and renamed the directory accordingly.

fiboknacky:
Based heavily on AddHotelAd.

### DIFF
--- a/examples/HotelAndTravel/AddHotelAd.php
+++ b/examples/HotelAndTravel/AddHotelAd.php
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-namespace Google\Ads\GoogleAds\Examples\HotelAds;
+namespace Google\Ads\GoogleAds\Examples\HotelAndTravel;
 
 require __DIR__ . '/../../vendor/autoload.php';
 
--- a/examples/HotelAndTravel/AddThingsToDoAd.php
+++ b/examples/HotelAndTravel/AddThingsToDoAd.php
@@ -1,0 +1,337 @@
+<?php
+
+/**
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Ads\GoogleAds\Examples\HotelAndTravel;
+
+require __DIR__ . '/../../vendor/autoload.php';
+
+use GetOpt\GetOpt;
+use Google\Ads\GoogleAds\Examples\Utils\ArgumentNames;
+use Google\Ads\GoogleAds\Examples\Utils\ArgumentParser;
+use Google\Ads\GoogleAds\Examples\Utils\Helper;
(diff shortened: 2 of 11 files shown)


examples/HotelAndTravel/AddThingsToDoAd.php
```diff
@@ -0,0 +1,337 @@
...
+     */
+    private static function addCampaignBudget(GoogleAdsClient $googleAdsClient, int $customerId)
+    {
+        // Creates a campaign budget.
+        $budget = new CampaignBudget([
+            'name' => 'Interplanetary Cruise Budget #' . Helper::getPrintableDatetime(),
+            'delivery_method' => BudgetDeliveryMethod::STANDARD,
+            // Sets the amount of budget.
+            'amount_micros' => 50000000,
+            // Makes the budget explicitly shared.
```
jradcliff, line 161: Is this a requirement for Things to do campaigns?

examples/HotelAndTravel/AddThingsToDoAd.php
```diff
@@ -0,0 +1,337 @@
...
+        string $campaignResourceName
+    ) {
+        // Creates an ad group.
+        $adGroup = new AdGroup([
+            'name' => 'Earth to Mars Cruise #' . Helper::getPrintableDatetime(),
+            // Sets the campaign.
+            'campaign' => $campaignResourceName,
+            // Sets the ad group type to TRAVEL_ADS. This cannot be set to other types.
+            'type' => AdGroupType::TRAVEL_ADS,
+            'cpc_bid_micros' => 10000000,
```
jradcliff, line 270: Just double-checking: does setting this when the bidding strategy is MaximizeConversionValue make sense? Is it just capping the max bid?

examples/HotelAndTravel/AddThingsToDoAd.php
```diff
@@ -0,0 +1,337 @@
...
+     * @param string $adGroupResourceName the resource name of ad group that a new ad group ad will
+     *     belong to
+     */
+    // [START add_things_to_do_ad_3]
+    private static function addAdGroupAd(
+        GoogleAdsClient $googleAdsClient,
+        int $customerId,
+        string $adGroupResourceName
+    ) {
+        // Creates a new ad group ad and set a travel ad info.
```
jradcliff, line 308: nit: 'set' => 'sets'

examples/HotelAndTravel/AddHotelAd.php
```diff
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-namespace Google\Ads\GoogleAds\Examples\HotelAds;
+namespace Google\Ads\GoogleAds\Examples\HotelAndTravel;
```
jradcliff, line 19: WDYT of just `Travel` instead?

examples/HotelAndTravel/AddThingsToDoAd.php
fiboknacky: Yes. Confirmed.

examples/HotelAndTravel/AddThingsToDoAd.php
fiboknacky: Thanks for catching. I think it makes sense to add it here either. The code works without this.

examples/HotelAndTravel/AddHotelAd.php
fiboknacky: That's fine too. I just checked and it seems we use Travel to cover both Hotel and Travel now.

examples/HotelAndTravel/AddThingsToDoAd.php
jradcliff: Could you add a comment so readers know `false` is not an option?

fiboknacky: Done.

jradcliff (approved):

Raibaz (approved):

PR #911, opened 2023-05-31
feat: even more resource helpers

noahdietz:


damatmedya:
Hello to all users,
When I try to use this library, I get a 500 error and I cannot use it. I provided all the installations but could not run it.
How can I simply get the campaign list and the balance and statistics used

fiboknacky:
Could you please create a new issue with all the details (e.g., the step you used to install this library) instead?

PR #922, opened 2023-06-29
Added a check for grpc versions when building GoogleAdsClient

fiboknacky:
The grpc versions installed using PECL and Composer can sometimes not match. This results in a bug that is hard to debug sometimes.

### DIFF
--- a/src/Google/Ads/GoogleAds/Lib/V14/GoogleAdsClientBuilder.php
+++ b/src/Google/Ads/GoogleAds/Lib/V14/GoogleAdsClientBuilder.php
@@ -31,6 +31,7 @@ use Grpc\Interceptor;
 use InvalidArgumentException;
 use Psr\Log\LoggerInterface;
 use Psr\Log\LogLevel;
+use UnexpectedValueException;
 
 /**
  * Builds Google Ads API clients.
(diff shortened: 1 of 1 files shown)


src/Google/Ads/GoogleAds/Lib/V14/GoogleAdsClientBuilder.php
```diff
@@ -415,6 +416,48 @@ public function validate()
                 'The gRPC channel credential can only be set when the transport is "grpc".'
             );
         }
+        // Check if the version of the grpc extension installed by Composer is greater than that of
+        // the extension installed by PECL, throw an exception to remind the user to upgrade.
+        $grpcPackageVersion = floatval(phpversion('grpc'));
```
Raibaz, line 421: Using `floatval` here confused me a bit because I didn't know how it worked with versions with more than a dot in it, e.g. 1.1.1.
I looked at [the reference doc for it](https://www.php.net/manual/en/function.floatval.php) and found out it only considers the last dot, so version 1.1.1 would be parsed as 11.1.
Since this is the case, would you think using `intval` rather than `floatval` here and on line 457 would achieve the same behavior and generate less confusion in a reader?

(I know, this is a very picky nit 🤣 )

Raibaz (approved):
LGTM pending a very small nit

src/Google/Ads/GoogleAds/Lib/V14/GoogleAdsClientBuilder.php
fiboknacky: `intval` wouldn't work because the version difference is mostly at the second or third digit.

From what I tested, 1.1.1 and 1.1.2 would be converted to 1.1, so it does no harm here.
But I think we can just convert them to string and let the string comparison do the work instead, which should be more accurate.

WDYT?

src/Google/Ads/GoogleAds/Lib/V14/GoogleAdsClientBuilder.php
```diff
@@ -415,6 +416,48 @@ public function validate()
...
+            throw new UnexpectedValueException(
+                'The grpc extension installed by Composer has a greater version than that'
+                . ' installed by PECL. Upgrade the PECL extension to avoid issues caused by the'
+                . ' version difference. For linux, run "sudo pecl install grpc".'
+            );
+        }
+    }
+
+    /**
+     * Gets the grpc version installed by Composer from the composer.lock file. Returns null if the
```
jradcliff, line 433: `Returns null if the if that information` ->
`Returns null if that information`

src/Google/Ads/GoogleAds/Lib/V14/GoogleAdsClientBuilder.php
```diff
@@ -31,6 +31,7 @@
 use InvalidArgumentException;
 use Psr\Log\LoggerInterface;
 use Psr\Log\LogLevel;
+use UnexpectedValueException;
```
jradcliff, line 34: Would it be feasible to add a test for this?

src/Google/Ads/GoogleAds/Lib/V14/GoogleAdsClientBuilder.php
fiboknacky: Good catch. version_compare is probably the best here.

Raibaz (approved):

jradcliff (approved):

src/Google/Ads/GoogleAds/Lib/V14/GoogleAdsClientBuilder.php
```diff
@@ -415,6 +416,51 @@ public function validate()
                 'The gRPC channel credential can only be set when the transport is "grpc".'
             );
         }
+        // Check if the version of the grpc extension installed by Composer is greater than that of
+        // the extension installed by PECL, throw an exception to remind the user to upgrade.
+        $grpcPackageVersion = phpversion('grpc');
+        $grpcComposerVersion = $this->getGrpcComposerVersion();
+        if (
+            !empty($grpcComposerVersion)
```
kwolniak, line 424: what if `$grpcPackageVersion` is `false` and `$grpcComposerVersion` has some value?

fiboknacky: Nice catch. I think another condition `!empty($grpcPackageVersion)` is missing.

kwolniak: exactly, in my setup where I don't have the extension installed it throws an exception :(

fiboknacky: Sorry for your inconvenience. :( Will create a PR soon.

PR #928, opened 2023-07-24
Fixed the case when system package grpc version not found

fiboknacky:
Also, added unit tests for the check added in #922.

To do so, I separated the version fetching logic into a class named `Dependencies`, so we can do dependency injection to `GoogleAdsClientBuilder`.

### DIFF
--- a/src/Google/Ads/GoogleAds/Util/Dependencies.php
+++ b/src/Google/Ads/GoogleAds/Util/Dependencies.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * Copyright 2023 Google LLC
  *
--- a/src/Google/Ads/GoogleAds/Util/Dependencies.php
+++ b/src/Google/Ads/GoogleAds/Util/Dependencies.php
@@ -1,0 +1,67 @@
+<?php
+/**
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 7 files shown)


src/Google/Ads/GoogleAds/Util/Dependencies.php
```diff
@@ -0,0 +1,67 @@
...
+class Dependencies
+{
+    /**
+     * Gets the grpc version installed by Composer from the specified file (default to
+     *  `composer.lock`. Returns null if that information cannot be found by any causes.
+     *
+     * @param string $fileName the file name to extract the grpc version from
+     * @return null|string the grpc version installed by Composer
+     */
+    public function getGrpcComposerVersion(string $fileName = 'composer.lock'): ?string
```
jradcliff, line 32: Tests are the only code that will ever specify a different filename from the default, correct?

jradcliff (approved):

tests/Google/Ads/GoogleAds/Util/DependenciesTest.php
```diff
@@ -0,0 +1,64 @@
+<?php
+
+/*
+ * Copyright 2020 Google LLC
```
PierrickVoulet, line 4: nit: 2023

tests/Google/Ads/GoogleAds/Util/DependenciesTestProvider.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+/*
+ * Copyright 2019 Google LLC
```
PierrickVoulet, line 4: nit: 2023

PierrickVoulet (changes requested):

PierrickVoulet (approved):

PR #999, opened 2024-03-08
Reverted back all PHP 8 features done in the following PRs to support legacy users

fiboknacky:
This is the revert of #941, #943, #944 with some slight brush-ups in some places.


Raibaz (approved):

sundquist (approved):
LGTM

Issue #971, opened 2023-11-29
Reminder: Updating the minimum required PHP version for this library to PHP 8.1 


fiboknacky:
By March 31, 2024, the composer file of this library will require PHP version 8.1 or later.

PHP 8.0 was sunset in November 26, 2023. The PHP development team no longer provides security fixes for the versions, so we highly recommend migrating from this version as soon as possible.

mjainavas:
Is there any patch to run [google-ads-php-v21.1.0](https://github.com/googleads/google-ads-php/releases/tag/v21.1.0) on PHP 7.4 ?
Its bit impossible at the moment to upgrade server, and Google Ads API V13 is about to sunset in next 10 days.

fiboknacky:
Unfortunately, we have no plan to patch the old version for older PHP versions.

mjainavas:
Thanks for update,
Is PHP 8.0 version good for V21.1.0, Can I consider this with LTS?

fiboknacky:
We don't patch the old release when we sunset an old PHP binary version, so v21.1.0 will continue to work with PHP 8 until the last API version (v15) is sunset.

fiboknacky:
Done with PR #1002

PR #1002, opened 2024-03-26
Update composer.json to require PHP 8.1

fiboknacky:
This is an update as planned in #971.


Raibaz (approved):
I'm guessing the kokoro failure is related to the other CL you sent, so re-running it now should make it pass.

LGTM anyway.

Raibaz (approved):

PR #1035, opened 2024-06-26
Added support for REST HTTP handler.

fiboknacky:


Raibaz (approved):

src/Google/Ads/GoogleAds/Lib/ConfigurationTrait.php
```diff
@@ -207,4 +210,14 @@ public function getGrpcInterceptors()
     {
         return $this->grpcInterceptors;
     }
+
+    /**
+     * Gets the REST HTTP handler.
+     *
+     * @return Guzzle6HttpHandler|Guzzle7HttpHandler
+     */
+    public function getRestHttpHandler()
```
bshaffer, line 219: The term `Rest` already implies `Http`, and. vice-versa, so why not call this simply `getHttpHandler`?

src/Google/Ads/GoogleAds/Lib/ConfigurationTrait.php
```diff
@@ -47,6 +49,7 @@ trait ConfigurationTrait
     private $unaryMiddlewares;
     private $streamingMiddlewares;
     private $grpcInterceptors;
+    private $restHttpHandler;
```
bshaffer, line 52: same as above - why not just call this `$httpHandler`?

bshaffer (commented):
nit for naming, otherwise LGTM

src/Google/Ads/GoogleAds/Lib/ConfigurationTrait.php
fiboknacky: Done.

src/Google/Ads/GoogleAds/Lib/ConfigurationTrait.php
fiboknacky: Done.

bshaffer (approved):

bshaffer:
FYI there's also a [credentialsConfig option for `authHttpHandler`](https://github.com/googleapis/gax-php/blob/28aa3e95969a75b278606a88448992a6396a119e/src/CredentialsWrapper.php#L103), which you could set as well (or add as a separate option). This is used in OAuth2 requests if there's an HTTP request to the OAuth2 "token" endpoint (or anywhere else, such as in external account credentials.  This is set on `credentialsConfig`:


```php
$clientoptions = ['credentialsConfig' => ['authHttpHandler' => $myHttpHandler]];

```

You could set this to the `httpHandler` above as well, when `null`, for consistency. Or you can add your own `setAuthHttpHandler` method for it, or do nothing at all. But I wanted to let you know it exists in case that's relevant!

fiboknacky:
Thanks. That'd be another PR then.

psobhanlo:
```
    public function search(SearchGoogleAdsRequest $request, array $callOptions = []): PagedListResponse
    {
        return $this->startApiCall('Search', $request, $callOptions);
    }

```

```
            $response = $googleAdsClient->getGoogleAdsServiceClient()->search(
                $customerId,
                $query,
                [
                    'pageSize' => $entriesPerPage,
                    // Requests to return the total results count. This is necessary to
                    // determine how many pages of results exist.
                    'returnTotalResultsCount' => true,
                    // There is no need to go over the pages we already know the page tokens for.
                    // Fetches the last page we know the page token for so that we can retrieve the
                    // token of the page that comes after it.
                    'pageToken' => end($pageTokens)
                ]
            );
```

Issue #1056, opened 2024-12-05
Deprecation notices when using with PHP 8.4


mikkamp:
**Your client library and Google Ads API versions:**

* Client library version: `dev-legacy-v25.0.0`
* Google Ads API version: `V18`
**Your environment:**
PHP 7. 4 up to PHP 8.4

**Description of the bug:**
When using the library on PHP 8.4 there are several deprecation notices which are logged. They all seem to be coming from the same function `parseName` which must make the `$template` parameter type nullable.

There might be other occurrences of PHP 8.4 incompatibilities in the library, but these are the ones we encountered from the parts that we use.

**Steps to reproduce:**
Use one of the services from the library with PHP 8.4 ensuring that deprecation notices are being logged.

**Expected behavior:**
No deprecation notices when running on PHP 8.4

**Request/Response Logs:**
These are the notices we are currently encountering:
```
GoogleAdsServiceClient::parseName(): Implicitly marking parameter $template as nullable is deprecated, the explicit nullable type must be used instead
google-listings-and-ads/vendor/googleads/google-ads-php/src/Google/Ads/GoogleAds/V16/Services/Client/GoogleAdsServiceClient.php:1789

CustomerServiceClient::parseName(): Implicitly marking parameter $template as nullable is deprecated, the explicit nullable type must be used instead
google-listings-and-ads/vendor/googleads/google-ads-php/src/Google/Ads/GoogleAds/V16/Services/Client/CustomerServiceClient.php:163

ConversionActionServiceClient::parseName(): Implicitly marking parameter $template as nullable is deprecated, the explicit nullable type must be used instead
google-listings-and-ads/vendor/googleads/google-ads-php/src/Google/Ads/GoogleAds/V16/Services/Client/ConversionActionServiceClient.php:157

CampaignServiceClient::parseName(): Implicitly marking parameter $template as nullable is deprecated, the explicit nullable type must be used instead
google-listings-and-ads/vendor/googleads/google-ads-php/src/Google/Ads/GoogleAds/V16/Services/Client/CampaignServiceClient.php:287

ProductLinkInvitationServiceClient::parseName(): Implicitly marking parameter $template as nullable is deprecated, the explicit nullable type must be used instead
google-listings-and-ads/vendor/googleads/google-ads-php/src/Google/Ads/GoogleAds/V16/Services/Client/ProductLinkInvitationServiceClient.php:164

[...]

PR #1059, opened 2024-12-12
Added support for PHP 8.4

fiboknacky:
- Regenerated source code to suppress the deprecation warnings.
- Updated src/Google/Ads/GoogleAds/Lib/InsecureCredentialsWrapper.php to be compatible with the latest version of GAX.

Bug: N/A


fiboknacky:
This is also the fix for https://github.com/googleads/google-ads-php/issues/1056.

Raibaz (approved):

PR #1109, opened 2025-10-09
PHP 8.4 Compat: Throw UnexpectedValueException for calls to undefined methods.

peterwilsoncc:
This is to cover a deprecation in PHP 8.4 `Passing E_USER_ERROR to trigger_error() is deprecated since 8.4`, per [the PHP 8.4 migration guide](https://www.php.net/manual/en/migration84.deprecated.php#:~:text=Using%20trigger_error()%20with%20E_USER_ERROR).

All calls to `trigger_error('Call to undefined method ' . __CLASS__ . "::$method()", E_USER_ERROR);` are replaced by throwing an exception `throw new \UnexpectedValueException('Call to undefined method ' . __CLASS__ . "::$method()");`

This will continue triggering a fatal error for the library. I'm happy to bikeshed the correct type of exception to throw here, I went back and forth on a simple `Error` but decided against it in the end.

Note, this comes out of a ticket for https://github.com/woocommerce/google-listings-and-ads/pull/3117, a WooCommerce/WordPress plugin making use of the `dev-legacy-v31.0.1` release. This is due to the need to support PHP 7.4.

If this change is acceptable to you all, it would be great if you could backport it to a v31 branch. I'm happy to create a pull request for this if you wish.

### DIFF
--- a/src/Google/Ads/GoogleAds/V19/Services/Client/AccountBudgetProposalServiceClient.php
+++ b/src/Google/Ads/GoogleAds/V19/Services/Client/AccountBudgetProposalServiceClient.php
@@ -261,7 +261,7 @@ class AccountBudgetProposalServiceClient
     public function __call($method, $args)
     {
         if (substr($method, -5) !== 'Async') {
-            trigger_error('Call to undefined method ' . __CLASS__ . "::$method()", E_USER_ERROR);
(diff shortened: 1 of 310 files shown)


ajs424:
Since there hasn't been any updates on this PR in quite some time I'll go ahead and close it out. If this is still a relevant concern with the current version /legacy branch of the PHP client library please feel free to reopen or create an updated PR.

PR #1114, opened 2025-11-05
Replace deprecated getLabel with isRepeated

ajs424:
Change-Id: Ia8d0307f98af5b4268de8c6629227f134eeb1755

### DIFF
--- a/src/Google/Ads/GoogleAds/Util/FieldMasks.php
+++ b/src/Google/Ads/GoogleAds/Util/FieldMasks.php
@@ -331,9 +331,13 @@ class FieldMasks
      */
     private static function isFieldRepeated(FieldDescriptor $fieldDescriptor)
     {
-        // The isRepeated() method is the recommended non-deprecated replacement 
-        // for checking FieldDescriptorProto\Label::LABEL_REPEATED.
-        return $fieldDescriptor->isRepeated();
+        // SAFE SOLUTION: Check if the modern method exists (for updated C extensions or native PHP).
+        // This protects against a fatal error in environments where the C extension is outdated.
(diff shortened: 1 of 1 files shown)


bshaffer:
I do not have access to approve the PR, but LGTM!

Raibaz (approved):

ajs424:
I believe the minimum PHP version required by the google/protobuf library has historically been well below PHP 7.4.

The google/protobuf packages in the 3.x series (including v3.26.0) are fully compatible with PHP versions 7.2 up to the latest 8.x versions. Since the composer file specifies a baseline requirement of "php": ">=7.3",  Protobuf version ^3.26.0 will work  running PHP 7.4. 

TLDR: This dependency upgrade does not drop support for PHP 7.4. The fix is strictly about modernizing the Protobuf API usage within the client library, which is compatible with the older PHP version still used by partners 

@fiboknacky @Raibaz

ajs424:
Hi all,

I wanted to clarify the dependency discussion regarding the required Protobuf version.

The Code Fix: The implementation now uses $fieldDescriptor->isRepeated().

The Dependency: Our main branch's composer.json already specifies:

`google/protobuf": "^3.21.5 || >=4.26 <=4.30.0`

Since this range already allows versions like v3.26.0 (which we determined supports isRepeated()), I have decided to keep the current composer.json file as is, with no change to the Protobuf constraint.

This preserves compatibility for any existing partners who are strictly locked to v3.21.5 on older branches, while still ensuring the fix works for users running a wider range of Protobuf versions.

Please let me know your thoughts of if you have any insight!

PR #1174, opened 2026-07-23
Relax composer constraints

ajs424:
Change-Id: Ic3b86f3bca611f6030ba6d3580c0f670d75f5c99


ajs424:
@bshaffer I relaxed the composer constraints here after testing with PIE. I ran some of the failed unit tests from the version mismatch and this seemed to resolve it. Any thoughts and/or concerns?

bshaffer:
This looks good to me, although I don't see relaxed constraints here, I see 

 - stricter grpc constraint
 - support for protobuf v5

